PR #1, opened 2023-09-02
Add evaluate service

juliangruber:
This has already been deployed to fly.io 🙏 

Closes https://github.com/filecoin-station/roadmap/issues/29
Closes https://github.com/filecoin-station/roadmap/issues/33

@bajtos happy to fulfil change requests or hand over

TODO:

- [ ] self heal / discover all measurements that haven't been preprocessed yet, and all rounds that haven't been evaluated yet. _This can also be added in a new PR_


index.js
```diff
@@ -0,0 +1,45 @@
...
+}) => {
+  const rounds = {}
+  const cidsSeen = []
+  const roundsSeen = []
+
+  // Listen for events
+  ieContract.on('MeasurementsAdded', (cid, _roundIndex) => {
+    const roundIndex = Number(_roundIndex)
+    if (cidsSeen.includes(cid)) return
+    cidsSeen.push(cid)
```
bajtos, line 18: IIUC, this array will keep growing until we run out of memory.

Have you considered remembering only the last N items, where N is for example 1000?

Also, what happens when we restart the service, e.g. after deploying a new version? The list of CIDs seen will be lost. How is that going to impact the evaluation?

index.js
```diff
@@ -0,0 +1,45 @@
...
+      roundIndex,
+      web3Storage,
+      logger
+    }).catch(console.error)
+  })
+
+  ieContract.on('RoundStart', _roundIndex => {
+    const roundIndex = Number(_roundIndex)
+    if (roundsSeen.includes(roundIndex)) return
+    roundsSeen.push(roundIndex)
```
bajtos, line 34: Ditto. 

Can we limit the size of this array? Is it perhaps enough to remember the last round seen?

What happens after the service restarts?

lib/evaluate.js
```diff
@@ -0,0 +1,49 @@
...
+const { BigNumber } = ethers
+
+export const evaluate = async ({
+  rounds,
+  roundIndex,
+  ieContractWithSigner,
+  logger
+}) => {
+  // Get measurements
+  const measurements = rounds[roundIndex] || []
```
bajtos, line 12: Since we are keeping all measurements in memory, I think we should handle the case where our service starts in the middle of an IE round, and we receive only some of the `MeasurementsAdded` events for this round. Unless that's something you will solve by implementing the self-healing feature? In that case, I am fine to defer that to a new pull request.

test/evaluate.js
```diff
@@ -0,0 +1,100 @@
...
+    const ieContractWithSigner = {
+      async setScores (roundIndex, peerIds, scores, summary) {
+        assert.strictEqual(roundIndex, 0)
+        assert.deepStrictEqual(peerIds, ['0x123'])
+        assert.strictEqual(scores.length, 1)
+        assert.strictEqual(
+          scores[0].toString(),
+          BigNumber.from(1_000_000_000_000_000).toString()
+        )
+        assert.match(summary, /^\d+ retrievals$/)
```
bajtos, line 22: Would you mind applying the same approach as I described in https://github.com/filecoin-station/spark-api/pull/63#discussion_r1314694193 and move the assertions to the end of the test after we call `evaluate()`?

The same comment applies to other similar tests in this PR too.

lib/preprocess.js
```diff
@@ -0,0 +1,42 @@
...
+  web3Storage,
+  logger
+}) => {
+  const measurements = (await fetchMeasurements(web3Storage, cid))
+    .filter(measurement => {
+      try {
+        assertValidMeasurement(measurement)
+        return true
+      } catch (err) {
+        logger.error('Invalid measurement', err.message, measurement)
```
bajtos, line 17: How can we gain visibility into this? Can we e.g. create a Grafana chart showing us the number (or fraction) of invalid measurements per committed batch or per SPARK round?

bajtos: It's probably best to improve observability in a follow-up PR though.

lib/preprocess.js
```diff
@@ -0,0 +1,42 @@
...
+    rounds[roundIndex] = []
+  }
+  rounds[roundIndex].push(...measurements)
+}
+
+const fetchMeasurements = async (web3Storage, cid) => {
+  const res = await web3Storage.get(cid)
+  const files = await res.files()
+  const measurements = JSON.parse(await files[0].text())
+  return measurements
```
bajtos, line 33: Let's discuss attack vectors.

- What happens when the CID points to a file that's not a JSON? Is our entire service going to crash on an unhandled error?
- What if `measurements` is not an array, or the array contain items that are not valid measurement objects?
- What if the first CID file is very large, e.g. several GB of JSON? Will it trigger denial-of-service?

bajtos, line 33: Observability (after this PR is landed): I would consider reporting the number of measurements as a new telemetry point and visualising it in Grafana. I think both the "number of measurements per CID" and "number of measurements per round" are potentially useful metrics to watch.

bajtos (commented):
Nice! I have a few minor comments to discuss.

> self heal / discover all measurements that haven't been preprocessed yet, and all rounds that haven't been evaluated yet.
> This can also be added in a new PR

Yeah, let's do that after this initial PR is landed.

index.js
juliangruber: Added a fix 👍 

The service restarting isn't a problem, we're mostly debouncing here anyway
- if we preprocess redundantly, there will be extra work but the results won't change
- if we evaluate redundantly, there will be extra work and gas, but the results won't change

index.js
juliangruber: See above

bajtos (approved):
LGTM as the first incremental version 👏🏻  

I consider my three comments above about observability and error handling as still relevant, but they can be addressed in follow-up pull requests.

lib/preprocess.js
juliangruber: https://github.com/filecoin-station/spark-evaluate/issues/2

lib/preprocess.js
juliangruber: https://github.com/filecoin-station/spark-evaluate/issues/2

juliangruber:
https://github.com/filecoin-station/spark-evaluate/issues/2

lib/evaluate.js
juliangruber: https://github.com/filecoin-station/spark-evaluate/issues/3

lib/preprocess.js
bajtos: > > What happens when the CID points to a file that's not a JSON? Is our entire service going to crash on an unhandled error?
> > What if measurements is not an array, or the array contain items that are not valid measurement objects?
> 
> These will only create a `console.error`, see `index.js`.

I see; this is good.

Can you please report these errors to Sentry as well? (I am ok to do that after this PR is landed if you prefer, but please create a follow-up issue so that we don't forget.)

> > What if the first CID file is very large, e.g. several GB of JSON? Will it trigger denial-of-service?
> 
> This is an attack vector 👍 It can make the service crash (it won't lead to inconsistent data or anything like that). In the meridian design where you run fraud detection on only a single measurement, we could make it so only one measurement ever is stored in memory. We could also use a streaming parser here.
> 
> In the current spark design, evaluation happens in memory, given all measurements of the round. Therefore I think as long as this is the case, we're not adding attack vectors at this part. Eventually we should make the service more stable by probably moving that calculation into postgres, or find some other way to not require looking at all of the data.

Can we reject CIDs where the data is larger than a reasonable threshold? If we expect our measurement service to commit 1000 CIDs at a time, then I think 1MB is a reasonable limit.

Such change does not prevent an attacker from committing many CIDs per round to cause the out-of-memory problem later in our evaluation pipeline, so maybe it's not worth implementing this measure on its own.

Let's open a follow-up GH issue where we can discuss different approaches and agree on what is meaningful to implement now & what we want to implement later.

juliangruber: > Can you please report these errors to Sentry as well? (I am ok to do that after this PR is landed if you prefer, but please create a follow-up issue so that we don't forget.)

https://github.com/filecoin-station/spark-evaluate/issues/4

> Can we reject CIDs where the data is larger than a reasonable threshold? If we expect our measurement service to commit 1000 CIDs at a time, then I think 1MB is a reasonable limit.

Such change does not prevent an attacker from committing many CIDs per round to cause the out-of-memory problem later in our evaluation pipeline, so maybe it's not worth implementing this measure on its own.

Let's open a follow-up GH issue where we can discuss different approaches and agree on what is meaningful to implement now & what we want to implement later.

Got it, can you please open that issue, as it's Spark specific?

PR #7, opened 2023-09-19
fix: setup console logger

bajtos:
Before my change, the evaluation step was failing with the following
error:

    TypeError: Cannot read properties of undefined (reading 'error')
      at file:///app/lib/preprocess.js:17:16
      at Array.filter (<anonymous>)
      at preprocess (file:///app/lib/preprocess.js:12:6)

In this change, I am setting up a console logger in the main entry point.

Connect to:
- https://github.com/filecoin-station/spark/issues/13


juliangruber (approved):
oops

PR #10, opened 2023-09-19
feat: rename peers to participants

bajtos:
- This is a follow-up for https://github.com/filecoin-station/spark-api/pull/89
- See also https://github.com/filecoin-station/spark/issues/13
- See also https://github.com/filecoin-station/spark-api/pull/97


lib/preprocess.js
```diff
@@ -9,6 +9,9 @@ export const preprocess = async ({
   logger
 }) => {
   const measurements = (await fetchMeasurements(web3Storage, cid))
+    // Rename "wallet_address" to "peerId"
```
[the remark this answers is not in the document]
juliangruber: yes please (see slack reply)

juliangruber (approved):

lib/preprocess.js
```diff
@@ -11,10 +12,12 @@ export const preprocess = async ({
   const measurements = (await fetchMeasurements(web3Storage, cid))
     .filter(measurement => {
       try {
+        // eslint-disable-next-line camelcase
+        measurement.participantAddress = parseParticipantAddress(measurement.wallet_address)
```
juliangruber, line 16: The callback of a `Array#filter` function shouldn't modify the data, otherwise the code is harder to reason about. Could you make it clearer by either mapping in an `Array#map` or refactoring this to a `for of` loop?

juliangruber (changes requested):

lib/preprocess.js
bajtos: > The callback of a `Array#filter` function shouldn't modify the data, otherwise the code is harder to reason about (one doesn't expect it).

Agreed, thanks for calling this out.

> Could you make it clearer by either mapping in an `Array#map` or refactoring this to a `for of` loop?

👍🏻

bajtos: Fixed in [7c976f7](https://github.com/filecoin-station/spark-evaluate/pull/10/commits/7c976f7e2e69cfaec821e2d45d6abe50fb255daa)

lib/preprocess.js
```diff
@@ -9,12 +10,27 @@ export const preprocess = async ({
   logger
 }) => {
   const measurements = (await fetchMeasurements(web3Storage, cid))
+    .map(measurement => {
+      try {
+        return {
+          ...measurement,
+          // eslint-disable-next-line camelcase
+          participantAddress: parseParticipantAddress(measurement.wallet_address)
```
juliangruber, line 18: isn't this logic duplicated now, once in measure and once in filter? and filter still changes the measurement

juliangruber (changes requested):

lib/preprocess.js
juliangruber: also, the measure service should submit `participant_address` instead of `wallet_address`. It's important that both services use the same terminology

bajtos: > isn't this logic duplicated now, once in measure and once in filter? and filter still changes the measurement

That was a mistake - I forgot to remove the code from the filter callback.

Fixed in 7c976f7e2e69cfaec821e2d45d6abe50fb255daa

> also, the measure service should submit `participant_address` instead of `wallet_address`. It's important that both services use the same terminology

In my view, SPARK modules knows nothing about participant addresses and FIL vs ETH formats. It receives the wallet address from the user and adds it to the measurement.

By keeping different field names (participant vs wallet), it's easier to reason about what kind of address is stored where.

- `wallet_address` is a Filecoin address (typically `f1...` or `f4...`)
- `participant_address` is an ETH address (`0x...`

juliangruber: The problem is that you can't just send a filecoin address to the contract, if the argument is of type `address` (which `setScores` has). See the Slack discussion for a suggestion

juliangruber: We decided against using FIL addresses, so that part of the contract can stay

juliangruber (approved):

bajtos:
I made a few more changes per our recent discussions on Slack, this PR is ready for review.

1. Renamed `evaluate` point field `total_peers` to `total_participants`
2. Changed the measurement parser to look for `participant_address` instead of `wallet_address`, see https://github.com/filecoin-station/spark-api/pull/97
3. Modified the code parsing the participant address to map f1/f2 addresses to the same hardcoded participant address starting with `0xf100`

PR #18, opened 2023-10-02
fix: handle missing `participant_address`

bajtos:
Before this change, we would reject measurements with no participant
address with an ugly error:

```
Invalid measurement: Cannot read properties of undefined (reading 'startsWith') {
```

In this commit, I changed the address parser to map measurements with
missing participant address to the same hard-coded `0xf100` participant
as we map `f1`/`t1` participants.

This is a follow-up for #10

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -59,7 +59,7 @@ const fetchMeasurements = async (web3Storage, cid) => {
 }
 
 export const parseParticipantAddress = filWalletAddress => {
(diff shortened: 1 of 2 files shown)


juliangruber (approved):

juliangruber (changes requested):
tests are failing 👋

bajtos:
🙈 

Fixed, PTAL again.

juliangruber (approved):

PR #19, opened 2023-10-02
test: remove delay after all tests are done

bajtos:
Before my change, there was a delay of approx. two seconds after all tests complete but before the Mocha process exits. This delay was caused by the recently added InfluxDB client (commit 09aaffcfe).

This commit modifies the test suite to explicitly close the InfluxDB client after all tests are finished.

### DIFF
--- a/lib/telemetry.js
+++ b/lib/telemetry.js
@@ -20,3 +20,5 @@ export const record = (name, fn) => {
   fn(point)
   writeClient.writePoint(point)
 }
+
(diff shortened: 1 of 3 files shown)


juliangruber (approved):
thank you!

bajtos:
This is pretty much the same thing I did in https://github.com/filecoin-station/spark-api/pull/98.

I added the `after` block to every test file so that the following use case is fixed too:

```
❯ mocha test/evaluate.js
```

PR #23, opened 2023-10-04
fix: don't record telemetry from tests

bajtos:
Modify the processing pipeline to accept the `record` function via dependency injection.

Rework the test suite to supply a dummy no-op `record` function so we don't corrupt telemetry data from production with telemetry submitted by tests.

Links to:
- https://github.com/filecoin-station/spark/issues/13
- https://github.com/filecoin-station/spark/issues/18

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ export const startEvaluate = ({
   ieContract,
   ieContractWithSigner,
   web3Storage,
-  record,
+  recordTelemetry,
   logger
 }) => {
   const rounds = {}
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -5,6 +5,7 @@ export const startEvaluate = ({
   ieContract,
   ieContractWithSigner,
   web3Storage,
+  record,
   logger
 }) => {
```
bajtos, line 10: I feel the name `record` is a bit ambiguous. Possible alternatives:
- `recordTelemetry`
- `recordTelemetryPoint`
- `telemetryPoint`

Thoughts?

juliangruber: I was on the fence with the name too. I like the ones starting with a verb like `record` and `track`, because the fits it being a function. Feel free to pick one 👍

juliangruber (approved):

PR #24, opened 2023-10-04
feat: reject measurements for unknown tasks

bajtos:
See https://github.com/filecoin-station/spark/issues/13

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -1,4 +1,5 @@
 import { record } from './telemetry.js'
+import { SPARK_API } from './config.js'
 
 export const evaluate = async ({
   rounds,
@@ -7,10 +8,12 @@ export const evaluate = async ({
   logger
 }) => {
   // Get measurements
+  /** @type {Record<string, any>[]} */
   const measurements = rounds[roundIndex] || []
 
   // Detect fraud
-  const honestMeasurements = measurements.filter(m => isHonestMeasurement(m))
+  await runFraudDetection(roundIndex, measurements)
+  const honestMeasurements = measurements.filter(m => m.fraudAssessment === 'OK')
 
   // Calculate reward shares
   const participants = {}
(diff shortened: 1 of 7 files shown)


test/evaluate.js
```diff
@@ -7,11 +7,24 @@ const { BigNumber } = ethers
...
+  participantAddress: VALID_PARTICIPANT_ADDRESS,
+  cid: 'QmUuEoBdjC8D1PfWZCc7JCSK8nj7TV6HbXWDHYHzZHCVGS',
+  provider_address: '/dns4/production-ipfs-peer.pinata.cloud/tcp/3000/ws/p2p/Qma8ddFEQWEU8ijWvdxXm3nxU7oHsRtCykAaVz8WUYhiKn',
+  protocol: 'bitswap'
+}
+
 describe('evaluate', () => {
   it('evaluates measurements', async () => {
-    const rounds = { 0: [] }
+    const roundIndex = ROUND_WITH_KNOWN_DETAILS
```
bajtos, line 24: Discussion point: in the current proposal, all tests are fetching details of a real round from the SPARK API endpoint.

The upside is that we know that all parts fit together well (this is integration-level testing).

The downside is ~300-700ms overhead for every test.

The alternative is to inject the function for getting details of the current round, similarly to how we inject `ieContractWithSigner` and `web3Storage`.

@juliangruber WDYT?

lib/typings.d.ts
```diff
@@ -0,0 +1,10 @@
+export interface RetrievalTask {
+  cid: String;
+  providerAddress: String;
+  protocol: String;
+}
+
+export interface RoundDetails {
+  roundId: string; // BigInt
```
juliangruber, line 8: can we use BigInts instead?

test/evaluate.js
juliangruber: +1 to injecting the function, this will also give more confidence in the tests because a flaky service won't cause issues

lib/evaluate.js
```diff
@@ -41,17 +44,64 @@ export const evaluate = async ({
...
   delete rounds[roundIndex]
 
+  // Calculate aggregates per fraud detection outcome
+  const assessments = {
+    OK: 0,
+    INVALID_TASK: 0
+  }
+  for (const m of measurements) {
+    assessments[m.fraudAssessment] = (assessments[m.fraudAssessment] ?? 0) + 1
+  }
```
juliangruber, line 54:
```suggestion
  // Calculate aggregates per fraud detection outcome
  const fraudAssessments = {
    OK: 0,
    INVALID_TASK: 0
  }
  for (const m of measurements) {
    fraudAssessments[m.fraudAssessment] = (fraudAssessments[m.fraudAssessment] ?? 0) + 1
  }
```

lib/evaluate.js
```diff
@@ -41,17 +44,64 @@ export const evaluate = async ({
...
   record('evaluate', point => {
     point.intField('round_index', roundIndex)
     point.intField('total_participants', Object.keys(participants).length)
     point.intField('total_measurements', measurements.length)
     point.intField('honest_measurements', honestMeasurements.length)
     point.intField('set_scores_duration_ms', setScoresDuration)
+
+    for (const a in assessments) {
+      point.intField(`measurements_${a}`, assessments[a])
+    }
```
juliangruber, line 65:
```suggestion
    for (const [type, count] of Object.entries(assessments)) {
      point.intField(`measurements_${type}`, count)
    }
```

To me this is simpler and more explicit

juliangruber (changes requested):

lib/typings.d.ts
bajtos: This interface describes the data we are getting from the IPFS. JSON does not support BigInt, therefore we are transmitting this value as a string.

test/evaluate.js
bajtos: Done in [302d589](https://github.com/filecoin-station/spark-evaluate/pull/24/commits/302d589d3945aeeb08e5fdcae97b1ae0bcf14351)

bajtos:
@juliangruber I reworked my PR to use `fetchRoundDetails` as an injectable dependency. It allowed me to revert many changes in the existing tests, which is a good sign.

The PR is ready for final review and landing. PTAL.

juliangruber (approved):

PR #28, opened 2023-10-11
test: add integration test for spark-api client

bajtos:
This is a follow-up for #24 where I forgot to commit this new file.


juliangruber (approved):
mighty fine

PR #29, opened 2023-10-11
feat: add debug logging

bajtos:


lib/evaluate.js
```diff
@@ -82,6 +86,8 @@ export const evaluate = async ({
 
 export const runFraudDetection = async (roundIndex, measurements, { fetchRoundDetails, recordTelemetry }) => {
   const roundDetails = await fetchRoundDetails(roundIndex, recordTelemetry)
+  debug(`ROUND DETAILS for round=${roundIndex}`, roundDetails)
```
juliangruber, line 89:
```suggestion
  debug(`ROUND DETAILS for round=%s`, roundIndex, roundDetails)
```

`debug` wants to use printf formatting, let's be consistent

lib/evaluate.js
```diff
@@ -1,3 +1,7 @@
+import debugFactory from 'debug'
+
+const debug = debugFactory('spark:evaluate')
```
juliangruber, line 3:
```suggestion
import createDebug from 'debug'

const debug = createDebug('spark:evaluate')
```

Not important, but I would prefer this naming

juliangruber (changes requested):

lib/evaluate.js
bajtos: I like that 👍🏻

bajtos:
@juliangruber I addressed both your comments, LGTY now?

lib/evaluate.js
```diff
@@ -82,6 +86,9 @@ export const evaluate = async ({
 
 export const runFraudDetection = async (roundIndex, measurements, { fetchRoundDetails, recordTelemetry }) => {
   const roundDetails = await fetchRoundDetails(roundIndex, recordTelemetry)
+  // IMPORTANT: The format string MUST NOT have any placeholder for the roundDetails object
```
juliangruber, line 89:
```suggestion
  // Omit the roundDetails object from the format string to get nicer formatting
```

Less screamy plus explains why

lib/evaluate.js
```diff
@@ -95,4 +102,17 @@ export const runFraudDetection = async (roundIndex, measurements, { fetchRoundDe
 
     m.fraudAssessment = 'OK'
   }
+
+  if (debug.enabled) {
+    for (const m of measurements) {
+      // Print round & participant address & CID together to simplify lookup when debugging
+      // IMPORTANT: The format string MUST NOT have any placeholder for the `m` object
```
juliangruber, line 109:
```suggestion
      // Omit the `m` object from the format string to get nicer formatting
```

juliangruber (approved):
Approved with tiny comment suggestions

PR #33, opened 2023-10-12
feat: replace Web3Storage with their HTTP GW

bajtos:
Rework the code fetching measurements from the IPFS network to use Web3Storage's public HTTP Gateway.

The primary motivation is to get rid of occasional errors reported by the Web3Storage client SDK.

The secondary motivation is to remove the dependency on Web3Storage secret (API access token).

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -6,12 +6,14 @@ export const preprocess = async ({
   rounds,
   cid,
   roundIndex,
-  web3Storage,
+  fetchMeasurements,
   recordTelemetry,
   logger
 }) => {
+  fetchMeasurements ??= fetchMeasurementsViaWeb3Gateway
+
   const start = new Date()
-  const measurements = await fetchMeasurements(web3Storage, cid)
+  const measurements = await fetchMeasurements(cid)
   const fetchDuration = new Date() - start
   const validMeasurements = measurements
     // eslint-disable-next-line camelcase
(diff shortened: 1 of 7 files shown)


lib/preprocess.js
```diff
@@ -56,10 +58,13 @@ export const preprocess = async ({
   })
 }
 
-const fetchMeasurements = async (web3Storage, cid) => {
-  const res = await web3Storage.get(cid)
-  const files = await res.files()
-  const measurements = JSON.parse(await files[0].text())
+export const fetchMeasurementsViaWeb3Gateway = async (cid) => {
+  const res = await fetch(`https://${encodeURIComponent(cid)}.ipfs.w3s.link/measurements.json`)
```
juliangruber, line 62: As discussed in https://filecoinproject.slack.com/archives/C027XP5BTGB/p1697108440660949?thread_ts=1697013561.303109&cid=C027XP5BTGB, we're now missing content validation / trusting the gateway. Are we ok with that, or shall we do as suggested in the thread?

juliangruber: Ah, now I saw your question.

I think we have these options:

1. Trust the gateway (bad)
2. Continue using their client, if we can live with the errors. How frequent are they?
3. Build our own verification code. How much work would this be?

bajtos: > Trust the gateway (bad)

I agree this bad.

> Continue using their client, if we can live with the errors. How frequent are they?

It depends. Sometimes we get errors for several MeasurementsAdded events in a row. Sometimes all works well.

Let me troubleshoot this some more.

I suspect that switching from their client to their GW may not solve the problem with low availability of web3.storage retrievals. I am concerned that both their client and their GW may be going through the same place that's occasionally failing.

Let's see.

> Build our own verification code. How much work would this be?

Probably not too much work if we can reuse their npm packages. But then there is also ongoing maintenance, so I'd rather avoid this route.

bajtos:
Closing in favour of https://github.com/filecoin-station/spark-evaluate/pull/34

lib/preprocess.js
bajtos: I opened two follow-up issues:
- https://github.com/filecoin-station/spark-evaluate/issues/35
- https://github.com/filecoin-station/spark-evaluate/issues/36

PR #42, opened 2023-10-18
spark api now lives at https://api.filspark.com/

juliangruber:


lib/config.js
```diff
@@ -1 +1 @@
-export const SPARK_API = process.env.SPARK_API ?? 'https://spark.fly.dev'
+export const SPARK_API = process.env.SPARK_API ?? 'https://app.filspark.com'
```
bajtos, line 1: Should the prefix be `app.` or `api.`?

juliangruber, line 1:
```suggestion
export const SPARK_API = process.env.SPARK_API ?? 'https://api.filspark.com'
```

bajtos (approved):
:shipit:

PR #53, opened 2023-11-01
feat: round scores to MAX_SCORE

### LINKED ISSUE #49: Always round up evaluation to MAX_SCORE
We are calculating individual participant scores as a fraction of the sum of honest jobs and then mapping this fraction to an integer in the `0..MAX_SCORE` range. When we sum up the final scores, we often get a slightly smaller value due to rounding errors.

For example:

```
EVALUTE ROUND 939: Invoking IE.setScores(); number of participants: 2, total score: 999999999999999n
```

The IE contract does not consider the round evaluation to be finished until the scores add up to MAX_SCORE. As a result, we end up with rounds with evaluations that have never finished.

To fix this, the evaluation service should add a rounding item to the scores, assigning the remainder to the burner address `0x000000000000000000000000000000000000dEaDz`.

The IE contract MUST recognise the burner address and ignore evaluation scores for this address when paying out rewards.

bajtos:
Ensure all scores add up to `MAX_SCORE` by adding a new entry assigning the remainder to the special address `0x000[...]000000000000000dEaD`.

The smart contract recognises this address and does not pay any rewards to it.

Resolve #49

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -2,6 +2,8 @@ import createDebug from 'debug'
 
 const debug = createDebug('spark:evaluate')
 
+export const MAX_SCORE = 1_000_000_000_000_000n
+
 export const evaluate = async ({
   rounds,
   roundIndex,
@@ -25,6 +27,7 @@ export const evaluate = async ({
 
   // Calculate reward shares
   const participants = {}
+  let sum = 0n
   for (const measurement of honestMeasurements) {
     if (!participants[measurement.participantAddress]) {
       participants[measurement.participantAddress] = 0n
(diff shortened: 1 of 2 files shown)


lib/evaluate.js
```diff
@@ -25,16 +27,26 @@ export const evaluate = async ({
...
-    participants[participantAddress] = participantTotal *
-      1_000_000_000_000_000n /
+    const score = participantTotal *
+      MAX_SCORE /
       BigInt(honestMeasurements.length)
+    participants[participantAddress] = score
+    sum += score
+  }
+
+  if (Object.keys(participants).length && sum < MAX_SCORE) {
```
bajtos, line 45: Discussion point: how do we want to handle empty rounds when nobody participated?

We have a test to verify that we call the smart contract with an empty array of participants. The check `Object.keys(participants).length` keeps the test passing.

But is that the desired behaviour?

bajtos: > What about empty rounds just set the full score to the burner address? This way the IE marks the round as completed, yet not funds are wasted


SGTM. I wasn't sure if the smart contract would support that.

Opened a PR: https://github.com/filecoin-station/spark-evaluate/pull/56

PR #58, opened 2023-11-09
feat: cache fetched measurements in dry-run tool

bajtos:
I am investigating why we get `undefined` fraud assessments in our evaluations. To speed up the dry-run evaluations, I want to cache measurements to speed up sub-sequent dry-run evaluations of the same round.

### DIFF
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -1,6 +1,9 @@
 import { evaluate } from '../lib/evaluate.js'
 import { preprocess, fetchMeasurementsViaGateway } from '../lib/preprocess.js'
 import { fetchRoundDetails } from '../lib/spark-api.js'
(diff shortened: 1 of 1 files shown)


bin/dry-run.js
```diff
@@ -28,7 +31,22 @@ if (!measurementCids.length) {
   process.exit(1)
 }
 
+const cacheDir = fileURLToPath(new URL('../.cache', import.meta.url))
```
bajtos, line 34: FWIW, this directory is already on the `.gitignore` list.

https://github.com/filecoin-station/spark-evaluate/blob/ddfe10abd91eb440f1e5f0608b6778a737f15b81/.gitignore#L105

juliangruber (approved):

PR #61, opened 2023-11-16
feat: report unique tasks in honest measurements

bajtos:
Now that our CID dataset has ~300k items, there is little value in observing the count of samples. Instead, I would like to see how many unique retrieval tasks we are able to test each round and over time.

In this change, I am adding a new field to the existing telemetry point "evaluate" with the count of unique tasks found in the honest measurements.

While we could get this data by querying the `measurements` table, such query is extremely expensive and cannot filter out measurements not considered as honest.

Links:
- https://github.com/filecoin-station/roadmap/issues/43


juliangruber (approved):
👏

bajtos:
Visualisation:

Screenshot 2023-11-28 at 10 33 55

PR #66, opened 2023-11-22
test: update spark-api to use a more recent round

bajtos:
The endpoint `/rounds/meridian/contract/index` does not support historic rounds.

The spark-api change:
- https://github.com/filecoin-station/spark-api/pull/147

### DIFF
--- a/lib/spark-api.js
+++ b/lib/spark-api.js
@@ -12,6 +12,7 @@ export const fetchRoundDetails = async (
 ) => {
   const start = new Date()
   let status = 0
+  let taskCount
 
   try {
     const res = await fetch(`${SPARK_API}/rounds/meridian/${contractAddress}/${roundIndex}`)
@@ -22,6 +23,7 @@ export const fetchRoundDetails = async (
     }
     /** @type {import('./typings.d.ts').RoundDetails} */
     const details = res.json()
+    taskCount = details.retrievalTasks?.length
     return details
   } finally {
     recordTelemetry('fetch_tasks_for_round', point => {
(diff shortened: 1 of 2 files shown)


bajtos:
@juliangruber instead of changing this test, I can also update spark-api to bring back the slower lookup as a fall-back for rounds that don't have their meridian columns set.

https://github.com/filecoin-station/spark-api/pull/147/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346L171

I think it's not worth the effort, but let me know if you think it is.

bajtos:
Hmm, there is something wrong with the new implementation of "get current round details". Each test run fails with a different error message, we are getting a different number of tasks every time.

bajtos:
> Hmm, there is something wrong with the new implementation of "get current round details". Each test run fails with a different error message, we are getting a different number of tasks every time.

I am trying to fix that issue here: https://github.com/filecoin-station/spark-api/pull/149

bajtos:
I ended up purging the entire cache in the Cloudflare dashboard. I hope that fixes the problem.

bajtos:

bajtos:
> Hmm, there is something wrong with the new implementation of "get current round details". Each test run fails with a different error message, we are getting a different number of tasks every time.

It seems this is fixed now. I added a visualisation to Grafana to use the new telemetry field introduced by this PR, and all seems good.

Screenshot 2023-11-22 at 18 09 33

juliangruber (commented):
👍

PR #70, opened 2023-11-24
Update web3.storage API

juliangruber:
- [x] Matching `spark-api` PR https://github.com/filecoin-station/spark-api/pull/151

Files are now published to the root entry of the CID (vs at `/measurements.json`), therefore this PR and the `spark-api` PR need to be merged together.

Links:
- https://github.com/filecoin-station/spark-api/issues/141

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Evaluate service
 ## Development
 
 ```bash
-$ WALLET_SEED=$(cat secrets/mnemonic) WEB3_STORAGE_API_TOKEN=$(cat secrets/web3storage) npm start
+$ WALLET_SEED=$(cat secrets/mnemonic) npm start
 $ npm test
 ```
 
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -1,5 +1,5 @@
 import { evaluate } from '../lib/evaluate.js'
-import { preprocess, fetchMeasurementsViaGateway } from '../lib/preprocess.js'
+import { preprocess, fetchMeasurementsTrusted } from '../lib/preprocess.js'
 import { fetchRoundDetails } from '../lib/spark-api.js'
 import { readFile, writeFile, mkdir } from 'node:fs/promises'
 import path from 'node:path'
@@ -43,7 +43,7 @@ const fetchMeasurements = async (cid) => {
     if (err.code !== 'ENOENT') console.warn('Cannot read cached measurements:', err)
   }
 
-  const measurements = await fetchMeasurementsViaGateway(cid)
+  const measurements = await fetchMeasurementsTrusted(cid)
   await writeFile(pathOfCachedResponse, JSON.stringify(measurements))
   return measurements
 }
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -1,5 +1,5 @@
 import { evaluate } from '../lib/evaluate.js'
-import { preprocess, fetchMeasurementsTrustless } from '../lib/preprocess.js'
+import { preprocess, fetchMeasurements } from '../lib/preprocess.js'
 import { fetchRoundDetails } from '../lib/spark-api.js'
 import { readFile, writeFile, mkdir } from 'node:fs/promises'
 import path from 'node:path'
@@ -35,7 +35,7 @@ const cacheDir = fileURLToPath(new URL('../.cache', import.meta.url))
 await mkdir(cacheDir, { recursive: true })
 
 const recordTelemetry = (measurementName, fn) => { /* no-op */ }
-const fetchMeasurements = async (cid) => {
+const fetchMeasurementsWithCache = async (cid) => {
   const pathOfCachedResponse = path.join(cacheDir, cid + '.json')
   try {
     return JSON.parse(await readFile(pathOfCachedResponse, 'utf-8'))
@@ -43,7 +43,7 @@ const fetchMeasurements = async (cid) => {
     if (err.code !== 'ENOENT') console.warn('Cannot read cached measurements:', err)
   }
 
-  const measurements = await fetchMeasurementsTrustless(cid)
+  const measurements = await fetchMeasurements(cid)
   await writeFile(pathOfCachedResponse, JSON.stringify(measurements))
   return measurements
 }
@@ -58,7 +58,7 @@ for (const cid of measurementCids) {
     roundIndex,
     rounds,
     cid,
-    fetchMeasurements,
+    fetchMeasurements: fetchMeasurementsWithCache,
     recordTelemetry,
     logger: console
   })
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -1,5 +1,5 @@
 import { evaluate } from '../lib/evaluate.js'
-import { preprocess, fetchMeasurementsTrusted } from '../lib/preprocess.js'
+import { preprocess, fetchMeasurementsTrustless } from '../lib/preprocess.js'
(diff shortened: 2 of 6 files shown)


lib/preprocess.js
```diff
@@ -89,15 +92,37 @@ export const preprocess = async ({
...
+    throw new Error(msg)
+  }
+  const reader = await CarReader.fromIterable(res.body)
+  const entries = exporter(cid, {
+    async get (cid) {
+      const block = await reader.get(cid)
+      await validateBlock(block)
+      return block.bytes
+    }
+  })
```
bajtos, line 110: Awesome that we can stream the content from a gateway and verify it in a streaming way 👏🏻

bin/dry-run.js
```diff
@@ -43,7 +43,7 @@ const fetchMeasurements = async (cid) => {
     if (err.code !== 'ENOENT') console.warn('Cannot read cached measurements:', err)
   }
 
-  const measurements = await fetchMeasurementsViaGateway(cid)
+  const measurements = await fetchMeasurementsTrusted(cid)
```
bajtos, line 46: Can we use `fetchMeasurementsTrustless` here too?

I added `fetchMeasurementsViaGateway` to allow people (including myself) to run this code without WEB3_STORAGE_API_TOKEN, and I was too lazy to implement streaming CAR validation. Now that we have the validation, I think there is no reason for not using it.

lib/preprocess.js
```diff
@@ -89,15 +92,37 @@ export const preprocess = async ({
...
+  for await (const entry of entries) {
+    if (entry.type === 'file' || entry.type === 'raw') {
```
bajtos, line 112: > Files are now published to the root entry of the CID (vs at /measurements.json), therefore this PR and the spark-api PR need to be merged together.

What are the implication of this for the upgrade path and for the evaluation of the round when we deploy these changes? Is it a lot of work to add support for `/measurements.json`?

bajtos, line 112: I find this suspicious. Under which conditions will we receive `file`, and when will we receive `raw` instead? Isn't it enough to have only one condition?

bajtos (approved):
I don't see any obvious problem. I have a few questions and discussion points, but none of that is blocking this PR from being merged & deployed.

bajtos:
Also, can you please grep through the source code and remove any remaining places that may still be using `WEB3_STORAGE_API_TOKEN`?

bin/spark-evaluate.js
```diff
@@ -47,7 +47,7 @@ const ieContract = new ethers.Contract(
 )
 const ieContractWithSigner = ieContract.connect(signer)
 const web3Storage = new Web3Storage({ token: WEB3_STORAGE_API_TOKEN })
-const fetchMeasurements = (cid) => fetchMeasurementsViaClient(web3Storage, cid)
+const fetchMeasurements = (cid) => fetchMeasurementsTrustless(web3Storage, cid)
```
bajtos, line 50: This looks like a bug to me, fetchMeasurementsTrustless accepts a single argument:

```js
export const fetchMeasurementsTrustless = async cid => {
```

Can you please add some tests for this function?

bajtos (changes requested):
I take that back - I think there is a bug.

bin/dry-run.js
juliangruber: Awesome, yes let's do that!

lib/preprocess.js
juliangruber: This came up from code review. Depending on the file size of the entry, it will either be plain text or binary encoded:

https://github.com/web3-storage/w3up/blob/e8bffe2ee0d3a59a977d2c4b7efe425699424e19/packages/upload-client/src/unixfs.js#L11

juliangruber: When we deploy a new `spark-evaluate` instance we sacrifice that round anyway, due to the in-memory storage. I think therefore this is fine. In the future once we use sqlite we need to think about this more

bin/spark-evaluate.js
juliangruber: Just spotted it as well. How do you propose we test this? This is the binary entry point 🤔

bajtos: > How do you propose we test this? This is the binary entry point 🤔

Ah, you are right. We haven't figured out how to automatically test our binary entry points yet, and I guess we have bigger fish to fry.

How about testing this manually - run the service locally (via npm start?), wait for the first measurement to be committed on the chain, and check it's correctly fetched.

README.md
```diff
@@ -7,7 +7,7 @@ Evaluate service
 ## Development
 
 ```bash
-$ WALLET_SEED=$(cat secrets/mnemonic) WEB3_STORAGE_API_TOKEN=$(cat secrets/web3storage) npm start
+$ WALLET_SEED=$(cat secrets/mnemonic) npm start
```
bajtos, line 10: 😍

bin/spark-evaluate.js
```diff
@@ -46,8 +43,7 @@ const ieContract = new ethers.Contract(
   provider
 )
 const ieContractWithSigner = ieContract.connect(signer)
-const web3Storage = new Web3Storage({ token: WEB3_STORAGE_API_TOKEN })
-const fetchMeasurements = (cid) => fetchMeasurementsViaClient(web3Storage, cid)
+const fetchMeasurements = (cid) => fetchMeasurementsTrustless(cid)
```
bajtos, line 46:
```suggestion
const fetchMeasurements = fetchMeasurementsTrustless
```

Or even better:

```js
startEvaluate({
  ieContract,
  ieContractWithSigner,
  fetchMeasurements: fetchMeasurementsTrustless,
  fetchRoundDetails,
  recordTelemetry,
  logger: console
})

File: lib/preprocess.js [hunk 7]
```diff
@@ -89,21 +92,33 @@ export const preprocess = async ({
...
-export const fetchMeasurementsViaClient = async (web3Storage, cid) => {
-  const res = await web3Storage.get(cid)
-  const files = await res.files()
-  const measurements = JSON.parse(await files[0].text())
-  return measurements
-}
-
-export const fetchMeasurementsViaGateway = async (cid) => {
-  const res = await fetch(`https://${encodeURIComponent(cid)}.ipfs.w3s.link/measurements.json`)
+export const fetchMeasurementsTrustless = async cid => {
```
bajtos, line 95: Now that we have only one implementation for fetching measurements, I propose to rename the function back to `fetchMeasurements`.

lib/preprocess.js
```diff
@@ -89,21 +92,33 @@ export const preprocess = async ({
...
+      return block.bytes
+    }
+  })
+  for await (const entry of entries) {
+    if (entry.type === 'file' || entry.type === 'raw') {
+      const bufs = []
+      for await (const buf of entry.content()) {
+        bufs.push(buf)
+      }
+      const json = Buffer.concat(bufs).toString()
```
bajtos, line 117: I think this could be implemented in a more performant way by using [`bl`](https://www.npmjs.com/package/bl), but I guess we are not that much performance/memory-usage sensitive here.

```suggestion
      const bufs = new BufferList()
      for await (const buf of entry.content()) {
        bufs.append(buf)
      }
      const json = bufs.toString()
```

bajtos (approved):
LGTM with few suggestions to consider.

bin/spark-evaluate.js
juliangruber: I guess Typescript is the most exciting option for bin script validation. Will run manually though for now

PR #71, opened 2023-11-27
fix: re-enable `group_winning_{min,max,mean}`

bajtos:
See https://github.com/filecoin-station/spark-evaluate/commit/ddfe10abd91eb440f1e5f0608b6778a737f15b81

### DIFF
--- a/test/evaluate.js
+++ b/test/evaluate.js
@@ -87,18 +87,10 @@ describe('evaluate', () => {
     let point = telemetry.find(p => p.name === 'evaluate')
     assert(!!point,
       `No telemetry point "evaluate" was recorded. Actual points: ${JSON.stringify(telemetry.map(p => p.name))}`)
-    assert.strictEqual(point.fields.unique_tasks, '0i')
-    assert.strictEqual(point.fields.measurements, '0i')
-    // no more fields are set for empty rounds
-    assert.deepStrictEqual(Object.keys(point.fields), [
-      'round_index',
-      'measurements',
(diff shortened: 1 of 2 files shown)


test/evaluate.js
```diff
@@ -108,6 +108,10 @@ describe('evaluate', () => {
     assert(!!point,
       `No telemetry point "evaluate" was recorded. Actual points: ${JSON.stringify(telemetry.map(p => p.name))}`)
     assert.strictEqual(point.fields.unique_tasks, '0i')
+
+    assert.strictEqual(point.fields.group_winning_min, '1')
+    assert.strictEqual(point.fields.group_winning_mean, '1')
+    assert.strictEqual(point.fields.group_winning_max, '1')
```
bajtos, line 114: This is not the best test case because the rates are the same as for an empty round. I am kind of lazy to write a better one; let me know if you think it's worth adding a test.

bajtos:
I could not reproduce the bug, so this is my best-try attempt to re-enable this telemetry without causing any crashes.

juliangruber (approved):

test/evaluate.js
juliangruber: fine with me

bajtos:
BTW, it's seems that we are calculating winning rates correctly, here is the log message for the last round:

```
EVALUATE ROUND 523: Success rate of winning per-inet-group task reward:
{ min: 0.21022727272727273, max: 1, mean: 0.9638576475057185 }
```

PR #77, opened 2023-11-29
raise max heap to 2972MB

juliangruber:


bajtos (approved):
👌🏻 

It will be best to land this change around 15:10 if we want to minimise measurements lost during restart.

PR #84, opened 2023-12-06
Reduce preprocess memory usage

juliangruber:
- delete unused measurement properties
- add `Measurement` class

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -24,11 +24,15 @@ export const preprocess = async ({
     // eslint-disable-next-line camelcase
     .map(({ participant_address, ...measurement }) => {
       try {
-        return {
+        const m = {
           ...measurement,
           participantAddress: parseParticipantAddress(participant_address),
           retrievalResult: getRetrievalResult(measurement)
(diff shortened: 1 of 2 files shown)


lib/preprocess.js
```diff
@@ -8,6 +8,17 @@ import createDebug from 'debug'
 
 const debug = createDebug('spark:preprocess')
 
+export class Measurement {
+  constructor (m) {
+    for (const [key, value] of Object.entries(m)) {
+      if (['zinnia_version', 'attestation', 'participant_address'].includes(key)) continue
```
bajtos, line 14: Is it worth extracting this array into a top-level constant to save some time and memory allocations on re-creating the array for every call of the constructor?

bajtos:
> > How did you collect that information? Would you mind to modify dry-run.js always to print memory usage stats?
> 
> Added it to the dry-run script 🙏

I think you forgot to commit this part? I don't see any changes in `dry-run.js` in this PR.

juliangruber:
I had committed it to main

PR #85, opened 2023-12-06
store dates as numbers to save space

juliangruber:
before:

```
{
  rss: 919781376,
  heapTotal: 812875776,
  heapUsed: 768347440,
  external: 4985291,
  arrayBuffers: 176855
}
```

after optimization:

```
{
  rss: 854179840,
  heapTotal: 741507072,
  heapUsed: 702200464,
  external: 5007050,
  arrayBuffers: 198614
}
```

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -16,14 +16,21 @@ export class Measurement {
     this.spark_version = m.spark_version
     this.fraudAssessment = null
     this.inet_group = m.inet_group
-    this.finished_at = m.finished_at
+    this.finished_at = parseDateTime(m.finished_at)
     this.provider_address = m.provider_address
     this.protocol = m.protocol
     this.byte_length = m.byte_length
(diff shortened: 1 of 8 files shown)


lib/preprocess.js
```diff
@@ -16,16 +16,23 @@ export class Measurement {
...
+    this.start_at = parseDateTime(m.start_at)
+    this.first_byte_at = parseDateTime(m.first_byte_at)
+    this.end_at = parseDateTime(m.end_at)
   }
 }
 
+const parseDateTime = (str) => {
+  if (!str) return undefined
+  const value = new Date(str)
+  if (Number.isNaN(value.getTime())) return undefined
```
bajtos, line 32: Hmm, we should reject measurements with an invalid date. I think it's ok to throw an error here.

bajtos (approved):

PR #86, opened 2023-12-06
perf: cache string values

bajtos:
For each round, keep a lookup table with all known string values.  When creating a new Measurement instance, replace incoming string values with the cached string instances.

This reduces memory usage by another ~70%.

**Before:**
```
{
  rss: 893943808,
  heapTotal: 570556416,
  heapUsed: 540473960,
  external: 5228292,
  arrayBuffers: 76335
}
```

**After:**
```
{
  rss: 561840128,
  heapTotal: 458522624,
  heapUsed: 407380752,
  external: 5660805,
  arrayBuffers: 559664
}
```

**After pointerizing provider_address:**
```
{
  rss: 3752099840,
  heapTotal: 167313408,
  heapUsed: 161246552,
  external: 3754973060,
  arrayBuffers: 35375
}
```

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -9,16 +9,16 @@ import createDebug from 'debug'
 const debug = createDebug('spark:preprocess')
 
 export class Measurement {
-  constructor (m) {
-    this.participantAddress = parseParticipantAddress(m.participant_address)
-    this.retrievalResult = getRetrievalResult(m)
-    this.cid = m.cid
-    this.spark_version = m.spark_version
+  constructor (m, pointerize = (v) => v) {
+    this.participantAddress = pointerize(parseParticipantAddress(m.participant_address))
+    this.retrievalResult = pointerize(getRetrievalResult(m))
+    this.cid = pointerize(m.cid)
+    this.spark_version = pointerize(m.spark_version)
     this.fraudAssessment = null
-    this.inet_group = m.inet_group
+    this.inet_group = pointerize(m.inet_group)
     this.finished_at = m.finished_at
     this.provider_address = m.provider_address
-    this.protocol = m.protocol
+    this.protocol = pointerize(m.protocol)
     this.byte_length = m.byte_length
     this.start_at = m.start_at
     this.first_byte_at = m.first_byte_at
@@ -34,6 +34,21 @@ export const preprocess = async ({
   recordTelemetry,
   logger
 }) => {
+  if (!rounds[roundIndex]) {
+    rounds[roundIndex] = []
+    rounds[roundIndex]._strings = new Map()
+  }
+
+  const knownStrings = rounds[roundIndex]._strings
+  assert(knownStrings)
+  const pointerize = (str) => {
+    if (str === undefined || str === null) return str
+    const found = knownStrings.get(str)
+    if (found) return found
+    knownStrings.set(str, str)
+    return str
+  }
+
   const start = new Date()
   /** @type import('./typings').Measurement[] */
   const measurements = await fetchMeasurements(cid)
@@ -42,7 +57,7 @@ export const preprocess = async ({
     // eslint-disable-next-line camelcase
     .map(measurement => {
       try {
-        return new Measurement(measurement)
+        return new Measurement(measurement, pointerize)
       } catch (err) {
         logger.error('Invalid measurement:', err.message, measurement)
         return null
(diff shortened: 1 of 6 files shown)


lib/preprocess.js
```diff
@@ -34,6 +34,21 @@ export const preprocess = async ({
   recordTelemetry,
   logger
 }) => {
+  if (!rounds[roundIndex]) {
+    rounds[roundIndex] = []
+    rounds[roundIndex]._strings = new Map()
```
bajtos, line 39: I don't want to keep all strings forever. By attaching the lookup table to the `rounds` structure, we will automatically delete it after the round is evaluated.

Adding new properties to a built-in Array instance is a bit hacky. A nicer option is to refactor our code so that `rounds` store an object `{measurements, strings}` instead.

Is it worth it?

test/preprocess.js
```diff
@@ -100,6 +103,10 @@ describe('preprocess', () => {
   })
 })
 
+const stripStringsLookupTable = (rounds) => {
+  for (const r of Object.values(rounds)) delete r._strings
+}
```
juliangruber, line 108: Why do we need to do this here? Shouldn't it be unnoticeable from the outside?

lib/preprocess.js
juliangruber: Let's have it like this, since it's techy itself

lib/preprocess.js
```diff
@@ -34,6 +34,21 @@ export const preprocess = async ({
...
+  }
+
+  const knownStrings = rounds[roundIndex]._strings
+  assert(knownStrings)
+  const pointerize = (str) => {
+    if (str === undefined || str === null) return str
+    const found = knownStrings.get(str)
+    if (found) return found
```
juliangruber, line 47:
```suggestion
    if (knownStrings.has(str)) return knownStrings.get(str)
```

This way we're checking presence and not truthyness

juliangruber (changes requested):

test/preprocess.js
bajtos:
```
      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected ... Lines skipped

  {
    '0': [
...
        spark_version: undefined,
        start_at: undefined
+     },
+     _strings: Map(3) {
+       '0x000000000000000000000000000000000000dEaD' => '0x000000000000000000000000000000000000dEaD',
+       'UNKNOWN_ERROR' => 'UNKNOWN_ERROR',
+       'ig1' => 'ig1'
      }
    ]
  }
      + expected - actual
```

The refactoring described in https://github.com/filecoin-station/spark-evaluate/pull/86#discussion_r1417504143 would allow us to get the measurements without the strings lookup table.

juliangruber: can we define it as non iterable instead?

lib/preprocess.js
bajtos: How about this?

```suggestion
    const found = knownStrings.get(str)
    if (found !== undefined) return found
```

juliangruber: correctness > perf imo

bajtos: > correctness > perf imo

Of course!

But we never store `undefined` values in our Map. I believe my proposal is both correct and performant.

lib/preprocess.js
```diff
@@ -5,20 +5,21 @@ import { CarReader } from '@ipld/car'
 import { validateBlock } from '@web3-storage/car-block-validator'
 import { recursive as exporter } from 'ipfs-unixfs-exporter'
 import createDebug from 'debug'
+import { RoundData } from './round.js'
```
juliangruber, line 8: did you forget to push `round.js`? 🤔  I'm not seeing it in this PR

juliangruber (changes requested):

lib/preprocess.js
bajtos: Fixed, PTAL again.

lib/preprocess.js
```diff
@@ -5,20 +5,21 @@ import { CarReader } from '@ipld/car'
...
+    this.retrievalResult = pointerize(getRetrievalResult(m))
+    this.cid = pointerize(m.cid)
+    this.spark_version = pointerize(m.spark_version)
     this.fraudAssessment = null
-    this.inet_group = m.inet_group
+    this.inet_group = pointerize(m.inet_group)
     this.finished_at = parseDateTime(m.finished_at)
-    this.provider_address = m.provider_address
-    this.protocol = m.protocol
+    this.provider_address = pointerize(m.provider_address)
```
bajtos, line 21: Whoa! I forgot to pointerize this field in my original version. After adding this line, the memory usage is down to 30% of the previous version 🎉 🎊

PR #87, opened 2023-12-07
add parse NDJSON

juliangruber:
this is now able to both parse single JSON blobs as NDJSON strings

Deploy before https://github.com/filecoin-station/spark-api/pull/165

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -144,7 +144,7 @@ export const fetchMeasurements = async cid => {
         bufs.push(buf)
       }
       const json = Buffer.concat(bufs).toString()
-      return JSON.parse(json)
+      return json.split('\n').filter(Boolean).map(line => JSON.parse(line))
     }
   }
   throw new Error('No measurements found')
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -16,14 +16,21 @@ export class Measurement {
     this.spark_version = m.spark_version
     this.fraudAssessment = null
     this.inet_group = m.inet_group
-    this.finished_at = m.finished_at
+    this.finished_at = parseDateTime(m.finished_at)
     this.provider_address = m.provider_address
     this.protocol = m.protocol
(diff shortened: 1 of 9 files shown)


lib/preprocess.js
```diff
@@ -143,13 +143,18 @@ export const fetchMeasurements = async cid => {
       for await (const buf of entry.content()) {
         bufs.push(buf)
       }
-      const json = Buffer.concat(bufs).toString()
-      return JSON.parse(json)
+      return parseMeasurements(Buffer.concat(bufs).toString())
```
bajtos, line 146: Is it time to introduce `bl` module as discussed in https://github.com/filecoin-station/spark-evaluate/pull/70#discussion_r1408913253?

lib/preprocess.js
```diff
@@ -143,13 +143,18 @@ export const fetchMeasurements = async cid => {
...
     }
   }
   throw new Error('No measurements found')
 }
 
+export const parseMeasurements = str => {
+  const ret = str.split('\n').filter(Boolean).map(line => JSON.parse(line))
```
bajtos, line 153: This assumes the old JSON format was never pretty-printed, i.e. the array was on a single line.

I guess it's an okay assumption, but I think it's worth documenting it here. Could you please add a code comment?

lib/preprocess.js
```diff
@@ -143,13 +143,18 @@ export const fetchMeasurements = async cid => {
...
+  if (Array.isArray(ret[0])) return ret[0]
```
bajtos, line 154: Should we make this check more strict?

```suggestion
  if (ret.length === 1 && Array.isArray(ret[0])) return ret[0]
```

lib/preprocess.js
juliangruber: What are you basing this on? I think it's independent of this PR. I don't know if it would help, someone needs to measure and try

PR #88, opened 2023-12-07
feat: rework histogram calculations

bajtos:
Remove the module `hdr-histogram-js` which was using a lot of external memory and producing only approximage values.

Rework the histogram calculation to collect all numbers into an array, sort the array and then pick items at p10/p50/p90 index.

Memory usage before - notice we were using ~3.5GB external memory:
```
{
  rss: 3526049792,
  heapTotal: 164167680,
  heapUsed: 159240088,
  external: 3754973060,
  arrayBuffers: 35375
}
```

Memory usage after - external is down to 3MB, heap total is up by 90MB:
```
{
  rss: 375226368,
  heapTotal: 259784704,
  heapUsed: 195350720,
  external: 4083229,
  arrayBuffers: 35016
}
```

### DIFF
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -1,14 +1,6 @@
 import createDebug from 'debug'
-import * as hdr from 'hdr-histogram-js'
 
 const debug = createDebug('spark:retrieval-stats')
-
-await hdr.initWebAssembly()
-const createHistogram = () => hdr.build({
-  numberOfSignificantValueDigits: 5,
-  bitBucketSize: 'packed',
-  useWebAssembly: true
-})
 
 /**
  * @param {import('./typings').Measurement[]} measurements
@@ -43,9 +35,9 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   const inetGroups = new Set()
   let downloadBandwidth = 0
 
-  const ttfbHistogram = createHistogram()
-  const durationHistogram = createHistogram()
-  const sizeHistogram = createHistogram()
+  const ttfbValues = []
+  const durationValues = []
+  const sizeValues = []
 
   for (const m of measurements) {
     // `retrievalResult` should be always set by lib/preprocess.js, so we should never encounter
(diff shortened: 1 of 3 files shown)


bajtos:
RSS before: 3,526,049,792
RSS after: 375,226,368

If I am reading this correctly, our memory usage went down from 3.5 GB to 375 MB 💪🏻

lib/retrieval-stats.js
```diff
@@ -103,15 +90,18 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
...
+  const count = values.length
+  if (count < 1) return
+  values.sort()
+  point.intField(`${fieldNamePrefix}_min`, values[0])
+  point.intField(`${fieldNamePrefix}_mean`, values.reduce((sum, v) => sum + v, 0) / count)
+  point.intField(`${fieldNamePrefix}_max`, values[count - 1])
   for (const p of [10, 50, 90, 95]) {
-    point.intField(`${fieldNamePrefix}_p${p}`, histogram.getValueAtPercentile(p))
+    const ix = Math.floor(p * count / 100)
+    point.intField(`${fieldNamePrefix}_p${p}`, values[ix])
```
juliangruber, line 104: is this really a percentile? Eg with values `[0, 50, 100]` the 90th percentile should be `90` but with this logic would be `100`. See https://www.calculatorsoup.com/calculators/statistics/percentile-calculator.php

juliangruber (changes requested):
Massive improvement! I'm just concerned with correctness

lib/retrieval-stats.js
bajtos: Nice! Thank you for sharing a link to the precise calculation of indexes from percentiles.

We currently have more than 700k measurements; I think the rounding error in percentile calculation is negligible.

Let's land my PR as-is and then improve the `percentile to index` calculation in a follow-up pull request.

IMO, the memory savings are so huge that it's okay to have less precise stats for a while.

bajtos: I have already opened a PR, see here: https://github.com/filecoin-station/spark-evaluate/pull/89

PR #89, opened 2023-12-07
fix: calculating values at a given percentile

bajtos:
Rework the naive approach producing _discrete percentile_ to a different algorithm that produces _continuous percentile_.

Upside: we get more precise values.
Downside: the values are interpolated now.

Quoting from [Postgres docs](https://www.postgresql.org/docs/current/functions-aggregate.html)

> `percentile_cont`
> Computes the _continuous percentile_, a value corresponding to the specified _fraction_ within the ordered set of aggregated argument values. This will interpolate between adjacent input items if needed.
> 
> `percentile_disc`
> Computes the _discrete percentile_, the first value within the ordered set of aggregated argument values whose position in the ordering equals or exceeds the specified _fraction_. The aggregated argument must be of a sortable type.

This is a follow-up for #88

### DIFF
--- a/test/retrieval-stats.test.js
+++ b/test/retrieval-stats.test.js
@@ -1,6 +1,7 @@
+import assert from 'node:assert'
 import createDebug from 'debug'
 import { Point } from '../lib/telemetry.js'
-import { buildRetrievalStats } from '../lib/retrieval-stats.js'
+import { buildRetrievalStats, getValueAtPercentile } from '../lib/retrieval-stats.js'
 import { VALID_MEASUREMENT } from './helpers/test-data.js'
 import { assertPointFieldValue } from './helpers/assertions.js'
 
@@ -107,5 +108,14 @@ describe('retrieval statistics', () => {
     assertPointFieldValue(point, 'duration_p10', undefined)
     assertPointFieldValue(point, 'duration_mean', undefined)
     assertPointFieldValue(point, 'duration_p90', undefined)
+  })
+})
+
+describe('getValueAtPercentile', () => {
+  it('interpolates the values', () => {
+    assert.strictEqual(
+      getValueAtPercentile([10, 20, 30], 90),
+      28
+    )
   })
 })
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -104,7 +104,7 @@ const addHistogramToPoint = (point, fieldNamePrefix, values) => {
   }
 }
 
-const getValueAtPercentile = (values, p) => {
+export const getValueAtPercentile = (values, p) => {
   // See https://www.calculatorsoup.com/calculators/statistics/percentile-calculator.php
   // 1. Arrange n number of data points in ascending order: x1, x2, x3, ... xn
   const n = values.length


bajtos:
Discussion point: should I add some tests for `getValueAtPercentile()`?

test/retrieval-stats.test.js
```diff
@@ -66,15 +66,15 @@ describe('retrieval statistics', () => {
...
     assertPointFieldValue(point, 'duration_p10', '2000i')
     assertPointFieldValue(point, 'duration_mean', '18500i')
-    assertPointFieldValue(point, 'duration_p90', '50000i')
+    assertPointFieldValue(point, 'duration_p90', '41000i')
 
-    assertPointFieldValue(point, 'car_size_p10', '1024i')
+    assertPointFieldValue(point, 'car_size_p10', '1331i')
     assertPointFieldValue(point, 'car_size_mean', '52430080i')
-    assertPointFieldValue(point, 'car_size_p90', '209715200i')
+    assertPointFieldValue(point, 'car_size_p90', '146801254i')
```
juliangruber, line 77: in comparison to the original expensive approach, does this make our valuers more right or wrong now?

bajtos:
> @bajtos I changed the base so that we get a proper diff

Hmm, that did not work out well. Fixing...

Screenshot 2023-12-07 at 14 47 45

Screenshot 2023-12-07 at 14 47 59

test/retrieval-stats.test.js
bajtos: I think the new approach makes our values more right.

```
> getValueAtPercentile([10, 20, 30], 90)
28
```

bajtos: Before, we would pick `20` or `30` as the value at p90.

The downside is that 28 picked by the new algorithm was not reported anywhere in the data, it's a synthetic value 🤷🏻‍♂️

bajtos: Verified using https://www.calculatorsoup.com/calculators/statistics/percentile-calculator.php

Screenshot 2023-12-07 at 14 54 41

bajtos: OTOH:

https://goodcalculators.com/percentile-calculator/

Screenshot 2023-12-07 at 14 55 26

bajtos: Postgres offers two functions:

`percentile_cont`
Computes the _continuous percentile_, a value corresponding to the specified _fraction_ within the ordered set of aggregated argument values. This will interpolate between adjacent input items if needed.

`percentile_disc`
Computes the _discrete percentile_, the first value within the ordered set of aggregated argument values whose position in the ordering equals or exceeds the specified _fraction_. The aggregated argument must be of a sortable type.

I think it's up to us to decide whether we want to use _continuous_ or _discrete_ percentiles.

@patrickwoodhead do you have any thoughts on this?

patrickwoodhead: Given we likely have more than 100 values, do we need to worry about interpolation? Let's say we have the numbers 1 to 200 and we want the 1st percentile. We first get the number 1 and 2. IIUC the above definitions, the continuous approach then gives us 1.5. the discrete approach gives us 1.

Suppose our data set is instead 1,10,11,20,21,30,31,.... for 200 values. The the two 1st percentile outputs respectively are 5.5 and 1. I think 5.5 is a better answer here.

So unless the dataset is really not a good fit for continuous, I vote we go for continuous, even if it means we aren't returning a data point in the data set.

juliangruber (approved):
I'm fine with this, we can add more percentile tests when we next want to change that function. Good stuff!

PR #90, opened 2023-12-07
feat: more validation during preprocessing

bajtos:
- refactor: types for raw and parsed measurements
- feat: more validation in pre-process step

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -10,6 +10,11 @@ import { RoundData } from './round.js'
 const debug = createDebug('spark:preprocess')
 
 export class Measurement {
+  /**
+   *
+   * @param {import('./typings.js').RawMeasurement} m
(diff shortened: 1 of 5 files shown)


lib/preprocess.js
```diff
@@ -10,27 +10,58 @@ import { RoundData } from './round.js'
...
+    }
+
+    if (endAt !== undefined && startAt === undefined) {
+      throw new Error('Invalid measurement: end_at is set, but start_at is not.')
+    }
+    this.duration = endAt - startAt
+    if (this.duration < 0) {
+      throw new Error('Invalid measurement: end_at is before start_at.')
+    }
   }
```
bajtos, line 56: TODO: check that `endAt > firstByteAt`

juliangruber (commented):
I'm all for validation as soon as possible, so that we get an early abort and clean data afterwards

PR #92, opened 2023-12-07
fix: retrieval stats (ttfb, car_size)

bajtos:
- fix: keep `status_code` in parsed measurements
- fix: ignore `byte_length` for non-200 status code

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -24,6 +24,7 @@ export class Measurement {
     this.start_at = parseDateTime(m.start_at)
     this.first_byte_at = parseDateTime(m.first_byte_at)
     this.end_at = parseDateTime(m.end_at)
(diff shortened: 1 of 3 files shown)


bajtos:
> fix: keep `status_code` in parsed measurements

This is needed to bring back `ttfb_*` metrics. We ignore measurements with non-200 status code when aggregating TTFB values. Before this change, all measurements had undefined status code and therefore did not contribute to TTFB data.

> fix: ignore `byte_length` for non-200 status code

When the retrieval fails (e.g. 404 or 502), spark-checker reports `byte_length: 0`. Before this change, we added these zeroes into our histogram, skewing percentiles towards zero.

juliangruber (approved):

juliangruber:
merging now to get more useful data over the weekend

PR #96, opened 2023-12-11
add pick random RPC provider on start (adds `chain.love`)

juliangruber:
Matches https://github.com/filecoin-station/spark-api/pull/169

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -1,8 +1,12 @@
 const {
   IE_CONTRACT_ADDRESS = '0xaaef78eaf86dcf34f275288752e892424dda9341',
-  RPC_URL = 'https://api.node.glif.io/rpc/v0',
+  RPC_URLS = 'https://api.node.glif.io/rpc/v0,https://api.chain.love/rpc/v1',
   SPARK_API = 'https://api.filspark.com'
 } = process.env
+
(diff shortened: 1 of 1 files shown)


juliangruber:
hindsight please

bajtos (commented):
LGTM.

This looks like a nice opportunity to extract a tiny helper module into a npm package.

juliangruber:
For selecting a random rpc url?

bajtos:
> > This looks like a nice opportunity to extract a tiny helper module into a npm package.
> For selecting a random rpc url?

Yeah, with sensible defaults for FEVM. So that people new to FEVM smart contracts don't have to learn about the different RPC API providers out there.

```js
const RPC_URL = getRandomJsonRpcProvider()
console.log(`Selected JSON-RPC endpoint ${RPC_URL}`)
```

juliangruber:
What do you mean by "sensible defaults for FEVM"?

I think you need to learn about RPC API providers. Not all support all methods for example, so one provider could work for project A but not project B

bajtos:
> What do you mean by "sensible defaults for FEVM"?
> 
> I think a user needs to learn about RPC API providers. Not all support all methods for example, so one provider could work for project A but not project B

I see. In that case, my proposal does not make sense.

PR #99, opened 2023-12-12
feat: add more percentiles to histogram stats

bajtos:
Add p1, p5 and p99 to all fields where we are collecting percentile values.


bajtos:

PR #97, opened 2023-12-11
feat: record subnets/participants/instances per task

bajtos:
Help us understand how many data points we have for each task (what is the effective committee size). Are the committees always large enough to give us enough confidence in finding an honest majority?

Also, record how many measurements (retrieval checks) we provide for each task to understand how much load we put on SPs.

InfluxDB point name: `committees`
Field prefixes:
- `subnets_`
- `participants_`
- `instances_`
- `measurements_` (all measurements, independent of fraud assessment)

Links:
- https://github.com/filecoin-station/spark-api/pull/171
- https://github.com/filecoin-station/spark/issues/45
- https://github.com/filecoin-station/roadmap/issues/52

PR #100, opened 2023-12-13
feat: report nodes participating in a round

bajtos:
- test: handle maxTasksPerNode in round details
- feat: report ~~instances~~ nodes participating in a round

Count the number of unique `(inet_group, participant_address)` nodes participating in the evaluated round and record it as another telemetry field.

Point name: `evaluate`
Field name: `total_nodes`

**IMPORTANT: the number of nodes calculated this way WILL BE DIFFERENT from the number of Station instances obtained from "ping" points.**

Related:
- https://github.com/filecoin-station/spark-evaluate/pull/97

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -116,7 +116,7 @@ export const evaluate = async ({
     point.intField('round_index', roundIndex)
     point.intField('total_participants', Object.keys(participants).length)
     point.intField('total_measurements', measurements.length)
-    point.intField('total_instances', countStationInstances(measurements))
+    point.intField('total_nodes', countUniqueNodes(measurements))
     point.intField('honest_measurements', honestMeasurements.length)
     point.intField('set_scores_duration_ms', setScoresDuration)
     point.intField('fraud_detection_duration_ms', fraudDetectionDuration)
@@ -357,7 +357,7 @@ const calculateInetGroupSuccessRates = (taskGroups) => {
 /**
  * @param {import('./typings').Measurement[]}
  */
-const countStationInstances = (measurements) => {
+const countUniqueNodes = (measurements) => {
   const instances = new Set()
   for (const m of measurements) {
     const key = `${m.inet_group}::${m.participantAddress}`
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -358,10 +358,10 @@ const calculateInetGroupSuccessRates = (taskGroups) => {
  * @param {import('./typings').Measurement[]}
  */
 const countUniqueNodes = (measurements) => {
-  const instances = new Set()
+  const nodes = new Set()
   for (const m of measurements) {
     const key = `${m.inet_group}::${m.participantAddress}`
-    instances.add(key)
+    nodes.add(key)
   }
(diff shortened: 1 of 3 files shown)


bajtos:
Next year, when I work on improving tasking & committee forming, I will need each spark node to have an identity - a public key for signing the messages. Once that's implemented, I can use this pub key as the instance identifier and rework how we compute telemetry while keeping the same point and field names in InfluxDB.

lib/evaluate.js
```diff
@@ -115,6 +115,7 @@ export const evaluate = async ({
     point.intField('round_index', roundIndex)
     point.intField('total_participants', Object.keys(participants).length)
     point.intField('total_measurements', measurements.length)
+    point.intField('total_instances', countStationInstances(measurements))
```
juliangruber, line 118: Can we please unify the naming? Currently we use `node` in the spec and `instances` in this PR. Or do they mean something different?

juliangruber (changes requested):

lib/evaluate.js
bajtos: Hmm, I see what you mean!

Which name do you prefer to use going forward - `node` or `instance`?

I am leaning towards using "node", because that's what we use in our design docs.

Switching to "node" also requires less changes. Changing "maxTasksPerNode" would require us to deploy a new version of SPARK.

The small downside is that https://github.com/filecoin-station/spark-evaluate/pull/97 already introduced the name "instance", so I need to rework that part + update Grafana charts.

juliangruber: Yeah I agree, `node` is nicer, especially since it's already in the spec

bajtos:
> The small downside is that https://github.com/filecoin-station/spark-evaluate/pull/97 already introduced the name "instance", so I need to rework that part + update Grafana charts.

See #102

juliangruber:
> Next year, when I work on improving tasking & committee forming, I will need each spark node to have an identity - a public key for signing the messages. Once that's implemented, I can use this pub key as the instance identifier and rework how we compute telemetry while keeping the same point and field names in InfluxDB.

Interesting! I assume every node will generate the private key on startup and cache it on disk? But, a fraudulent node could generate new keypairs in memory and use as many as they wish per machine

lib/evaluate.js
```diff
@@ -352,3 +353,15 @@ const calculateInetGroupSuccessRates = (taskGroups) => {
...
+/**
+ * @param {import('./typings').Measurement[]}
+ */
+const countUniqueNodes = (measurements) => {
+  const instances = new Set()
+  for (const m of measurements) {
+    const key = `${m.inet_group}::${m.participantAddress}`
+    instances.add(key)
+  }
+  return instances.size
```
juliangruber, line 366:
```suggestion
  const nodes = new Set()
  for (const m of measurements) {
    const key = `${m.inet_group}::${m.participantAddress}`
    nodes.add(key)
  }
  return nodes.size
```

juliangruber (changes requested):

juliangruber (approved):

PR #104, opened 2023-12-14
Add `tasks_per_node` retrieval stats. Closes #78

### LINKED ISSUE #78: Fraud detection: limit number of measurements per round
Detect when a checker node is submitting too many measurements.

In https://www.notion.so/pl-strflt/SPARK-tasking-v2-604e26d57f6b4892946525bcb3a77104, we are describing the following parameters:

- `CC` - “committee count” - how many committees we want to form. Initially, this will be a hardcoded parameter. We will manually tweak it over time based on the number of active checker nodes.
- `TC` - “tasks per committee” - how many tasks to define for each committee.
- `TN` - “tasks per node” - how many tasks a node performs. This value should be a fraction of TC, e.g. `TN = TC/5`

At the moment, `CC=1` and `TC=400`. 

Assuming retrievals are fast and can be completed in ~1 second, and the spark checker waits for 10 seconds between tasks, and the SPARK round is ~60 minutes, then an honest node can complete around 327 tasks per round.

Proposed changes:

- [x] Add a new fraud detection step to reward only the measurements for the first 300 unique tasks. Because measurements are not associated with Station instances, we should use `(participant_address, inet_group)` as the identifier.

juliangruber:
Closes #78

### DIFF
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -33,6 +33,7 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
 
   const participants = new Set()
   const inetGroups = new Set()
+  const tasksPerNode = new Map()
   let downloadBandwidth = 0
 
   const ttfbValues = []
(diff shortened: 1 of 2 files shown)


bajtos (approved):
👌🏻

PR #106, opened 2023-12-19
limit `setScores` arguments length

juliangruber:
In order not to run into smart contract argument length limitations


bajtos:
LGTM 👍🏻

A potentially missing test: call evaluate with max+1 participants and verify that we invoke the smart contract twice.

PR #115, opened 2024-01-09
refactor: use `just-percentile` instead of own impl

bajtos:
Refactor our code calculating percentiles to use a 3rd party dependency instead of our own implementation.

I discovered this while thinking how to address the discussion in my blog post:
https://github.com/bajtos/bajtos.github.io/pull/23#discussion_r1426613735

It makes me wonder why I didn't find this npm package before!

Sometimes, it's better to have more control over our code and not use 3rd-party dependencies for simple helper functions, so maybe this PR isn't an improvement.

Thoughts?

### DIFF
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -1,6 +1,11 @@
 import createDebug from 'debug'
+import getValueAtPercentile from 'just-percentile'
 
 const debug = createDebug('spark:retrieval-stats')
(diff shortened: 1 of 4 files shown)


juliangruber (approved):
I'm in favor of this chance, since it makes the codebase more manageable and it doesn't seem like we need more control over the implementation. We can always revert

PR #116, opened 2024-01-09
dry-run: cache list of events queried from chain

bajtos:
A small improvement to make sub-sequent dry runs faster to complete.


juliangruber (approved):
nice!

PR #117, opened 2024-01-10
cli: add `get-unique-values.js`

juliangruber:
I used this to get the unique participant addresses from last month:

```bash
$ DATABASE_URL=... node bin/get-unique-values.js participant_address 2023-12-09T00:00:00Z 2024-01-09T00:00:00Z
```


bajtos (approved):
Do you have any concerns about code duplication between the new script and "fetch-measurements"?

I find the name "get-unique-values" a bit confusing/vague, how about something like "get-unique-measurement-field-values" (that's awfully long 😢 ) or "get-unique-field-values"?

juliangruber:
I was thinking about deduplication as well, but don't think there's enough redundancy yet. Most code is just options parsing.

What about `get-unique-measurement-values`? For me, in a KV concept, you have keys and values, not keys and key values

bajtos:
> What about `get-unique-measurement-values`?

SGTM 👍🏻 

In physics, there are different things you can measure, and this process produces measurements. Like "speed measurement" or "electromagnetic field measurement". That brings the following script name to my mind: `get-unique-measurements`. (But that could be confusing because we already use the term "measurement" to describe the object with all fields, so one could interpret this name as fetching unique Measurement objects.)

Maybe we don't need to emphasise the uniqueness of values and call the script `get-measured-values`.

I am fine with any options discussed here, including the original one. Please pick one yourself.

>  For me, in a KV concept, you have keys and values, not keys and key values

I see, makes sense.

In my mind, this is not a KV store, but an Object database where we have objects (measurements) that have properties (with a name like "cid", "participant_address"), and these properties have values. So we have "property name" and "property value".

juliangruber:
Got it, I think `get-unique-measurement-property-values.js` is perfect then. Just long 😅 I'm going to keep the simple name for now, let's rename it in the future if I find it confusing (without the context of this discussion)

bajtos:
SGTM. I updated the PR title to match the final file name.

PR #121, opened 2024-01-11
preprocess: add cid to invalid block error

juliangruber:


lib/preprocess.js
```diff
@@ -134,7 +134,11 @@ export const fetchMeasurements = async cid => {
   const entries = exporter(cid, {
     async get (cid) {
       const block = await reader.get(cid)
-      await validateBlock(block)
+      try {
+        await validateBlock(block)
+      } catch (err) {
+        throw new Error(`Invalid block in ${cid}`, { cause: err })
+      }
```
bajtos, line 141: What if `reader.get` throws? Let's handle both cases together.

```js
try {
  const block = await reader.get(cid)
  await validateBlock(block)
} catch (err) {
 // ...
}
```

bajtos (approved):
Nice! Thank you for taking care of this.

I have one suggestion for you to consider.

lib/preprocess.js
juliangruber: If reader.get throws I think we should have a different error, since it's not about an invalid block. I think therefore once we want to debug this, we need to add a second try/catch

PR #122, opened 2024-01-11
feat: write public stats to our PG DB

bajtos:
A prototype implementing code for producing stats to power our Spark Public Dashboard. Design doc: 
https://pl-strflt.notion.site/DB-for-SPARK-public-data-a79cbb7a0ff5447980b714afdb2645d6

1. Add a runtime dependency on a PG database, using a new database
   name `spark_stats` (originally: ~~`spark_public`~~). Setup database schema migrations executed
   at service startup.

2. Define the first table `retrievals` to keep stats about all vs
   succesfull retrievals performed each day.

3. Update the evaluate step to update these stats for each round
   completed.

**TODO**

- [x] Setup PG DB for GitHub Actions to make the tests pass
- [x] Create a new PG user `spark_evaluate` and a new database `spark_stats`
- [x] Define DATABASE_URL in Fly.io secrets

Links:
- https://github.com/filecoin-station/roadmap/issues/57

### DIFF
--- a/bin/spark-evaluate.js
+++ b/bin/spark-evaluate.js
@@ -1,5 +1,5 @@
 import * as Sentry from '@sentry/node'
-import { IE_CONTRACT_ADDRESS, RPC_URL, rpcHeaders } from '../lib/config.js'
+import { DATABASE_URL, IE_CONTRACT_ADDRESS, RPC_URL, rpcHeaders } from '../lib/config.js'
 import { startEvaluate } from '../index.js'
 import { fetchRoundDetails } from '../lib/spark-api.js'
 import assert from 'node:assert'
@@ -9,6 +9,7 @@ import { newDelegatedEthAddress } from '@glif/filecoin-address'
 import { recordTelemetry } from '../lib/telemetry.js'
 import fs from 'node:fs/promises'
 import { fetchMeasurements } from '../lib/preprocess.js'
+import { migrateWithDbConfig } from '../lib/migrate.js'
 import http from 'node:http'
 
 const {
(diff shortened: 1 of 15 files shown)


bin/spark-evaluate.js
```diff
@@ -24,6 +24,8 @@ Sentry.init({
 
 assert(WALLET_SEED, 'WALLET_SEED required')
 
+await import('./migrate.js')
```
bajtos, line 27: I don't want to keep a long-running database connection since we write to the database only once per round (once per hour).

By importing the bin script, I don't have to write the code to open & close the PG connection.

migrations/001.do.retrieval-stats.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE retrievals (
```
juliangruber, line 1: `retrievals` is an ambiguous name, looks like here we're rather storing `retrieval_stats`

bin/spark-evaluate.js
juliangruber: I agree with the connection thinking, what about turning this into a lib then though?

bajtos: Done in e3d3762

migrations/001.do.retrieval-stats.sql
bajtos: Renamed in ca896e5

bajtos:
Last minute change: let's rename the DB from `spark_public` to `spark_stats`. It better captures what we are storing. In the future, we can store in this DB data for our private dashboards too.

juliangruber (approved):
Great work!

PR #124, opened 2024-01-11
fix log

juliangruber:


bajtos (approved):
Nice!

It makes me wonder what the type of `bucketIndex` is, considering that `buckets` is a regular array. 🤷🏻

PR #125, opened 2024-01-16
fix: setup a no-op PG client in dry-run

bajtos:
This is a follow-up for #122 to fix the dry-run script.

Links:
- https://github.com/filecoin-station/roadmap/issues/57

### DIFF
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -99,8 +99,12 @@ await evaluate({
   fetchRoundDetails,
   ieContractWithSigner,
   logger: console,
-  createPgClient,
-  recordTelemetry
+  recordTelemetry,
+
+  // We don't want dry runs to update data in `sparks_stats`, therefore we are passing a stub
+  // connection factory that creates no-op clients. This also keeps the setup simpler. The person
+  // executing a dry run does not need access to any Postgres instance.
+  createPgClient: createNoopPgClient
 })
 
 console.log(process.memoryUsage())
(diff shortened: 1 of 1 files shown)


bin/dry-run.js
```diff
@@ -209,3 +210,14 @@ async function fetchMeasurementsAddedFromChain (roundIndex) {
 
   return events.filter(e => e.roundIndex.eq(roundIndex)).map(e => e.cid)
 }
+
+function createPgClient () {
```
juliangruber, line 214: why is it ok to pass a mock here?

bajtos: I don't expect dry runs to modify any data. spark-evaluate uses the PG connection to update `spark_stats` database. I am using a mock because I don't want to do that. (And 

I also want a simpler setup that does not require the person running a dry run to have PG up & running and/or DATABASE_URL correctly configured.

juliangruber: Ok! Could you please add this as a comment here? Or, could we pass a function that mocks less? Assuming no queries will happen. I think it should throw if it tries to query the noop mock

bajtos: > Could you please add this as a comment here?

Done: 70ee0e4

> Or, could we pass a function that mocks less? Assuming no queries will happen. I think it should throw if it tries to query the noop mock

`evaluate()` is expected to run SQL queries to update the statistics. 

I am afraid we don't have a different function that we could execute for the dry run if we want the dry run to report all information produced by the real run - e.g. what scores we submit on the chain, what telemetry we report to InfluxDB and so on.

We are already passing a stub function for `recordTelemetry`. So at least, we are staying consistent in this PR.

bajtos: Sure, done in [7b451bb](https://github.com/filecoin-station/spark-evaluate/pull/125/commits/7b451bb4e13032a691d378e2ad164501fda19720).

PR #131, opened 2024-01-17
fix: typo when updating RSR stats

bajtos:


bajtos:
Is it worth adding a test for this? At the moment, we test only the path when a new row is inserted.

juliangruber:
> Is it worth adding a test for this? At the moment, we test only the path when a new row is inserted.

Since we had wrong data because of this, yeah 👍

Issue #79, opened 2023-11-29
More evaluation stats & visualisations & alerts


bajtos:
- [ ] How similar/dissimilar are measurements for the same task reported by different checker nodes?
- [ ] Round evaluation was not submitted within 5 minutes after a new round started
- [ ] Less than 10 participants scored in the round evaluation
- [ ] Daily & monthly active participants, correctly combining data from multiple rounds

PR #133, opened 2024-01-17
feat: report daily participants to public stats

bajtos:
Create a new `spark_stats` table to keep track of daily participants:

```sql
SELECT day::TEXT, COUNT(DISTINCT participant_id) as count
FROM daily_participants GROUP BY day;
```

This table allows us to correctly calculate monthly participants too:

```sql
SELECT
  date_trunc('month', day)::DATE::TEXT as month,
  COUNT(DISTINCT participant_id) as count
FROM daily_participants
GROUP BY month;
```

TODO:
- [x] add tests

Links:
- https://github.com/filecoin-station/spark-evaluate/issues/79

### DIFF
--- a/lib/public-stats.js
+++ b/lib/public-stats.js
@@ -21,14 +21,14 @@ export const updatePublicStats = async ({ createPgClient, honestMeasurements }) 
   const pgClient = await createPgClient()
   try {
     await updateRetrievalStats(pgClient, retrievalStats)
-    await updateDailyParticipants(pgClient, Array.from(participants.values()))
+    await updateDailyParticipants(pgClient, participants)
   } finally {
     await pgClient.end()
   }
 }
 
 /**
- * @param {import('./typings').CreatePgClient} createPgClient
+ * @param {import('pg').Client} pgClient
  * @param {object} stats
  * @param {number} stats.total
  * @param {number} stats.successful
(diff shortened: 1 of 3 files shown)


lib/public-stats.js
```diff
@@ -28,10 +44,48 @@ export const updatePublicStats = async ({ createPgClient, honestMeasurements })
...
+  ])
+}
+
+/**
+ * @param {import('./typings').CreatePgClient} createPgClient
+ * @param {string[]} participants
+ */
+const updateDailyParticipants = async (pgClient, participants) => {
+  debug('Updating daily participants (%s seen)', participants.length)
+  for (const participantAddress of participants) {
```
juliangruber, line 58: Are we ok with performing 2 SQL queries for every participant, in series? If it fails midway, will we have inconsistent data? Playing devil's advocate here

juliangruber (approved):
..otherwise LGTM!

lib/public-stats.js
bajtos: Great questions! 💯 

I'll think about this tomorrow.

bajtos: Yeah, I was focused so much on storage efficiency and the querying side, that I completely neglected the writing side.

With ~2k participants per round, my current implementation would run 4k SQL queries. That would take forever to complete.

Thanks for flagging this early! 😍

bajtos: Cool, I found a neat trick: We can use `SELECT UNNEST($1::TEXT[])` to build a query that accepts a JavaScript array parameter and converts the items into result rows.

Then I can change `INSERT...VALUES` to `INSERT...SELECT` to leverage this mechanism.

The changes were a bit more involved (see [7a2d14d](https://github.com/filecoin-station/spark-evaluate/pull/133/commits/7a2d14d11f10513e2946eba863f4721383ed322f)), but I expect the performance to be excellent.

juliangruber (approved):
Great work 😍

PR #135, opened 2024-01-19
Store only two rounds

juliangruber:
We only need the current round for preprocessing and the previous round for evaluation. This locks down memory growth of the rounds structure


index.js
```diff
@@ -28,10 +32,19 @@ export const startEvaluate = async ({
     cidsSeen.push(cid)
     if (cidsSeen.length > 1000) cidsSeen.shift()
 
+    if (!rounds.current) {
+      rounds.current = new RoundData(roundIndex)
+    } else if (rounds.current.index !== roundIndex) {
+      // This should never happen
+      throw new Error(
+        `Round index mismatch: ${rounds.current.index} !== ${roundIndex}`
+      )
```
juliangruber, line 41: I've added round validation logic, but am not sure this is worth it. This is just one commit, so easy to revert. Wdyt @bajtos?

juliangruber: I guess we should keep it if it's better for a measurement to be rejected than added to the wrong round.

lib/round.js
```diff
@@ -2,7 +2,8 @@ export class RoundData {
   /** @type {Map<string, string>} */
   #knownStrings
 
-  constructor () {
+  constructor (index) {
```
bajtos, line 5: Would you mind adding a jsdoc comment to describe the index type (is it BigInt?)?

I find the name index a bit ambiguous, roundIndex would be more descriptive, but it also feels a bit redundant to have "round.roundIndex", so I am fine to keep the current name.

lib/round.js
```diff
@@ -2,7 +2,8 @@ export class RoundData {
...
+    this.index = index
```
bajtos, line 6: AFAICT, we are not using this property anywhere - we always pass the round index in another variable/property alongside the round data. I think we don't need to introduce `this.index`, but it's also possible I have missed something.

WDYT?

test/evaluate.js
```diff
@@ -93,7 +92,7 @@ describe('evaluate', () => {
   })
 
   it('handles empty rounds', async () => {
-    const rounds = { 0: new RoundData() }
+    const round = new RoundData()
```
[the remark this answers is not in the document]
juliangruber: Changed it to always pass 0

lib/round.js
juliangruber: it's used in round index validation logic: https://github.com/filecoin-station/spark-evaluate/pull/135/commits/505ed331b017141ab0a9121ae41fc0b52e903232

bajtos: I see. Maybe we don't need to keep the index inside round data, but move it to a new variable in `index.js`, alongside `rounds`, `cidsSeen` and `roundsSeen`?

bajtos (approved):
LGTM with one more suggestion to consider:
https://github.com/filecoin-station/spark-evaluate/pull/135/files/73c276e5bc42f9f606cb579f6e111961d3f3c545..5253b2e909536abd02be9ba200bbdc3d170b3134#r1463356644

lib/round.js
juliangruber: What is the motivation for this change?

juliangruber: I'm going to merge for now, happy to follow up here

PR #144, opened 2024-02-01
feat: calculate stats for IPNI queries

bajtos:
Links:
- https://github.com/filecoin-station/spark/issues/40
- https://github.com/filecoin-station/spark-api/pull/204

PR #151, opened 2024-02-07
feat: new stat `rate_tasks_advertising_http`

bajtos:
Calculate the ratio of tasks where at least one measurement reported that HTTP retrieval was advertised to IPNI vs the number of tasks where at least measurement reported `indexerResult`.

This is a follow-up for https://github.com/filecoin-station/spark-evaluate/pull/144 which calculates the indexer result breakdown from all measurements, including measurements
submited by outdated Spark checkers that don't query IPNI at all.

Links:
- https://www.notion.so/pl-strflt/Moving-IPNI-queries-to-Spark-checkers-c58175c0f1c841e6bb82ac44a1472a98
- https://github.com/filecoin-station/spark/issues/40

### DIFF
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -21,6 +21,11 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
 
   const uniqueTasksCount = countUniqueTasks(measurements)
 
+  /** @type {Set<string>} */
+  const tasksAdvertisingHttp = new Set()
+  /** @type {Set<string>} */
+  const tasksQueryingIndexer = new Set()
+
   // Calculate aggregates per retrieval result
 
   // We are intentionally not initializing all possible keys here.
@@ -83,6 +88,10 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
 
     const node = `${m.inet_group}::${m.participantAddress}`
     tasksPerNode.set(node, (tasksPerNode.get(node) ?? 0) + 1)
+
+    const taskId = getTaskId(m)
+    if (m.indexerResult) tasksQueryingIndexer.add(taskId)
+    if (m.indexerResult === 'OK') tasksAdvertisingHttp.add(taskId)
   }
   const successRate = resultBreakdown.OK / totalCount
 
@@ -109,6 +118,11 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   for (const [result, count] of Object.entries(indexerResultBreakdown)) {
     telemetryPoint.floatField(`indexer_rate_${result}`, count / totalCount)
   }
+
+  telemetryPoint.floatField('rate_tasks_advertising_http', tasksQueryingIndexer.size > 0
+    ? tasksAdvertisingHttp.size / tasksQueryingIndexer.size
+    : 0
+  )
 }
 
 /**
@@ -137,14 +151,14 @@ const addHistogramToPoint = (point, values, fieldNamePrefix = '') => {
   }
 }
 
+// TODO: include Filecoin Deal id
+const getTaskId = (/** @type {import('./typings').Measurement} */m) => `${m.cid}`
+
 /**
  * @param {import('./typings').Measurement[]} measurements
  * @returns {number}
  */
 const countUniqueTasks = (measurements) => {
-  // TODO: include Filecoin Deal id
-  const getTaskId = (/** @type {import('./typings').Measurement} */m) => `${m.cid}`
-
   const uniqueTasks = new Set()
   for (const m of measurements) {
     const id = getTaskId(m)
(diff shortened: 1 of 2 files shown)


lib/retrieval-stats.js
```diff
@@ -83,6 +88,10 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
 
     const node = `${m.inet_group}::${m.participantAddress}`
     tasksPerNode.set(node, (tasksPerNode.get(node) ?? 0) + 1)
+
+    const taskId = getTaskId(m)
+    if (m.indexerResult) tasksQueryingIndexer.add(taskId)
```
juliangruber, line 93: what is a "task querying the indexer"? Or is it a task "from" querying the indexer?

lib/retrieval-stats.js
```diff
@@ -83,6 +88,10 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
...
+    if (m.indexerResult === 'OK') tasksAdvertisingHttp.add(taskId)
```
juliangruber, line 94: Also here, is the task really advertising http?

lib/retrieval-stats.js
```diff
@@ -109,6 +118,11 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   for (const [result, count] of Object.entries(indexerResultBreakdown)) {
     telemetryPoint.floatField(`indexer_rate_${result}`, count / totalCount)
   }
+
+  telemetryPoint.floatField('rate_tasks_advertising_http', tasksQueryingIndexer.size > 0
```
juliangruber, line 122:
```suggestion
  telemetryPoint.floatField('rate_of_tasks_advertising_http', tasksQueryingIndexer.size > 0
```

At first I interpreted "rate" as "to rate something"

juliangruber (changes requested):
code looks good to me, but naming is unclear

lib/retrieval-stats.js
bajtos: Let me explain. 

I want to calculate the answer to the following question: How many deals are advertising HTTP retrieval to IPNI?

I started with a naive approach = the fraction of measurements with `indexerResult: OK` from all measurements.

That does not work now because >95% of the measurements have node indexer results (they were submitted by old Spark checkers that don't query IPNI).

Also, we have a variable number of measurements for each task, ranging from ~7 to ~400. This distorts the calculated value, too.

To get a more precise value, I want to calculate the numbers for tasks, not measurements.

1. total = For how many tasks in this round do we have indexer query results?
2. http = For how many tasks in this round was there at least one indexer query result containing an HTTP retrieval advertisement?

The answer to my original question is then `http/total`.

I am using `tasksQueryingIndexer` to count `total`, and `tasksAdvertisingHttp` to count `http`.

I see how the current names can be confusing. Could you suggest better ones?

lib/retrieval-stats.js
bajtos: What would be a better name? How about `http_availability_rate`?

Alternatively: `tasks_with_http_retrieval_available`. (I find this rather long.)

juliangruber: What did you think of my suggestion?

lib/retrieval-stats.js
juliangruber: I think the confusing part is that a task isn't querying the indexer. What about

- `tasksWithIndexerResults`
- `tasksWithHTTPAdvertisement`

these are based on your descriptions

> total = For how many tasks in this round do we have indexer query results?
http = For how many tasks in this round was there at least one indexer query result containing an HTTP retrieval advertisement?

lib/retrieval-stats.js
bajtos: The name you suggested is a bit long, plus, as you wrote in the other thread, tasks are not advertising anything. 

So I guess a better name is `rate_of_tasks_with_http_advertisement`

I can also rename "task" to "deal", because that's where we want to arrive eventually - a task will be defined as "check if this deal is retrievable over HTTP?"


That gives us `rate_of_deals_advertising_http`, which sounds reasonable to me.

WDYT?

bajtos:
@juliangruber PTAL again

lib/retrieval-stats.js
juliangruber: good idea to move from tasks to deals. +1!

lib/retrieval-stats.js
```diff
@@ -137,14 +151,14 @@ const addHistogramToPoint = (point, values, fieldNamePrefix = '') => {
   }
 }
 
+// TODO: include Filecoin Deal id
+const getTaskId = (/** @type {import('./typings').Measurement} */m) => `${m.cid}`
```
juliangruber, line 155: should this be `getDealId` now?

bajtos: We are using the term "task" in many other places. To change this, I have to review those places and decide whether switching from "tasks" to "deals" makes sense.

Let's defer this until later, perhaps as part of adding deal/miner info to our tasks (https://github.com/filecoin-station/roadmap/issues/65).

PR #155, opened 2024-02-12
feat: collect IPNI query results for public stats

bajtos:
This PR will be followed by another to expose this new data via HTTP (see https://github.com/filecoin-station/spark-stats).

Links:
- https://www.notion.so/pl-strflt/Moving-IPNI-queries-to-Spark-checkers-c58175c0f1c841e6bb82ac44a1472a98
- https://github.com/filecoin-station/spark/issues/40
- https://github.com/filecoin-station/roadmap/issues/84

### DIFF
--- a/lib/public-stats.js
+++ b/lib/public-stats.js
@@ -1,6 +1,8 @@
 import assert from 'node:assert'
 import createDebug from 'debug'
 
+import { getTaskId } from './retrieval-stats.js'
+
 const debug = createDebug('spark:public-stats')
 
 /**
@@ -22,6 +24,7 @@ export const updatePublicStats = async ({ createPgClient, honestMeasurements }) 
   try {
     await updateRetrievalStats(pgClient, retrievalStats)
     await updateDailyParticipants(pgClient, participants)
+    await updateIndexerQueryStats(pgClient, honestMeasurements)
   } finally {
     await pgClient.end()
   }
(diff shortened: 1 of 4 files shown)


migrations/004.do.indexer-stats.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE indexer_query_stats (
+  day DATE NOT NULL PRIMARY KEY,
+  total INT NOT NULL,
+  advertising_http INT NOT NULL
```
bajtos, line 4: Perhaps `http_advertised` would be a better column name?

migrations/004.do.indexer-stats.sql
```diff
@@ -0,0 +1,5 @@
...
+);
```
bajtos, line 5: Do we want to keep track of the number of deals advertising Graphsync? On the one hand, it tells us how many SPs are advertising to IPNI (which is good) but not supporting HTTP yet (which may be easy to fix). OTOH, we don't really care about this metric, our ultimate goal is to have everybody advertising HTTP.

@patrickwoodhead thoughts?

migrations/004.do.indexer-stats.sql
juliangruber: I think we can improve this by looking at the topology. We have query stats, and for each day we're tracking..the total number of stats advertising http? (I'm being naive here). We are tracking the total number of tasks and tasks with http advertisement, etc, but the notion of a task is missing.

What do you think about changing the name from `indexer_query_stats.advertising_http` to `task_indexer_stats.advertising_http` or `task_indexing_stats.advertising_http`?

juliangruber (changes requested):

migrations/004.do.indexer-stats.sql
bajtos: We are tracking daily stats of how many tasks (deals tested) have HTTP retrieval advertised in IPNI.

I can rename `_stats` to `_daily_stats`, but it's a long name and we track all stats with daily granularity (e.g. `retrieval_stats`).

I can rename the columns to `tasks_total` and `tasks_advertising_http`, but isn't that too long and a bit redundant?

bajtos: Oh, and we also want to move from "tasks" to "deals" in the near future (see https://github.com/filecoin-station/roadmap/issues/65).

bajtos: That leads me to the following column names:

- `deals_tested`
- `deals_advertising_http`

bajtos: I quite like these two names, I applied the rename in 2ac16e1.

PTAL again

bajtos: 89c1762 updates also variable names

lib/public-stats.js
```diff
@@ -65,6 +68,35 @@ export const updateDailyParticipants = async (pgClient, participants) => {
...
 }
 
+const updateIndexerQueryStats = async (pgClient, honestMeasurements) => {
+  /** @type {Set<string>} */
+  const dealsWithHttpAdvertisement = new Set()
+  /** @type {Set<string>} */
+  const dealsWithIndexerResults = new Set()
+
+  for (const m of honestMeasurements) {
+    const dealId = getTaskId(m)
```
bajtos, line 78: The name `getTaskId` is not ideal, but also not that big of a deal IMO, so let's defer the cleanup/refactoring of this part until later (https://github.com/filecoin-station/roadmap/issues/65).

juliangruber (approved):

PR #159, opened 2024-02-20
fix: dry-run

bajtos:
Update the dry-run script to accommodate the recent changes we made: upgrade to Ethers v6, refactor per-round data, and so on.


juliangruber (approved):
👏

PR #176, opened 2024-03-25
Use more recent measurements for integration test

bajtos:
Pick measurements that include recently added `miner_id` and `provider_id` fields.

This pull request does not change any functionality, only the test. It allows
me to verify the impact of the next round of changes on round evaluation
results.

Links:
- https://github.com/filecoin-station/roadmap/issues/65

PR #177, opened 2024-03-25
feat: include minerId in task definition

bajtos:
When checking whether a measurement is for a valid round task, compare both `cid` and `minerId` fields

When picking a checker node in an IPv4 subnet to receive the reward, include `minerId` in the grouping key: When two nodes in the same subnet retrieve the same CID but from different miners, they should be both rewarded.

In practice, this second change will manifest very infrequently, because it's unprobable to have a single round with two tasks for the same CID.  As a result, I did not need to make any changes to the integration test running against real measurements.

Links:
- https://github.com/filecoin-station/spark-evaluate/pull/176
- https://github.com/filecoin-station/roadmap/issues/65

**Out of scope:**

Covered by future tasks in https://github.com/filecoin-station/roadmap/issues/65:

- Modify fraud detection (committee building) in spark-evaluate to cross-check ProviderID.
- Update spark-evaluate to calculate per-miner stats for each round and visualise these stats in our internal dashboards
- Measure and visualise requests per storage provider - https://github.com/filecoin-station/roadmap/issues/64

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -257,8 +257,9 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
     assert(typeof m.inet_group === 'string', 'missing inet_group')
     assert(typeof m.finished_at === 'number', 'missing finished_at')
 
-    // TODO: check also Filecoin Deal id
-    const isValidTask = sparkRoundDetails.retrievalTasks.some(t => t.cid === m.cid)
+    const isValidTask = sparkRoundDetails.retrievalTasks.some(
+      t => t.cid === m.cid && t.minerId === m.minerId
+    )
     if (!isValidTask) {
       m.fraudAssessment = 'INVALID_TASK'
     }
@@ -272,8 +273,7 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
   for (const m of measurements) {
     if (m.fraudAssessment) continue
 
-    // TODO: include Filecoin Deal id
-    const key = `${m.inet_group}::${m.cid}`
+    const key = `${m.inet_group}::${m.cid}::${m.minerId}`
     let group = taskGroups.get(key)
     if (!group) {
       group = []
@@ -341,10 +341,11 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
       // Print round & participant address & CID together to simplify lookup when debugging
       // Omit the `m` object from the format string to get nicer formatting
       debug(
-        'FRAUD ASSESSMENT for round=%s client=%s cid=%s',
+        'FRAUD ASSESSMENT for round=%s client=%s cid=%s minerId=%s',
         roundIndex,
         m.participantAddress,
         m.cid,
+        m.minerId,
         m)
     }
   }
@@ -370,9 +371,10 @@ const calculateInetGroupSuccessRates = (taskGroups) => {
 
   for (const groupMeasurements of taskGroups.values()) {
     debug(
-      '  measurements for inet_group=%s cid=%s',
+      '  measurements for inet_group=%s cid=%s minerId=%s',
       groupMeasurements[0].inet_group,
-      groupMeasurements[0].cid
+      groupMeasurements[0].cid,
+      groupMeasurements[0].minerId
     )
     // First, find participants that submitted some valid measurements for this task
     // and find out whether they were rewarded or not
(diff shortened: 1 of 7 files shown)


lib/evaluate.js
```diff
@@ -257,8 +257,9 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
     assert(typeof m.inet_group === 'string', 'missing inet_group')
     assert(typeof m.finished_at === 'number', 'missing finished_at')
 
-    // TODO: check also Filecoin Deal id
-    const isValidTask = sparkRoundDetails.retrievalTasks.some(t => t.cid === m.cid)
+    const isValidTask = sparkRoundDetails.retrievalTasks.some(
+      t => t.cid === m.cid && t.minerId === m.minerId
```
bajtos, line 261: FYI: I first updated a test to fail with an expected error and only then made this change.

lib/evaluate.js
```diff
@@ -272,8 +273,7 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
   for (const m of measurements) {
     if (m.fraudAssessment) continue
 
-    // TODO: include Filecoin Deal id
-    const key = `${m.inet_group}::${m.cid}`
+    const key = `${m.inet_group}::${m.cid}::${m.minerId}`
```
bajtos, line 276: FYI: I first updated a test to fail with an expected error and only then made this change.

lib/retrieval-stats.js
```diff
@@ -151,8 +151,7 @@ const addHistogramToPoint = (point, values, fieldNamePrefix = '') => {
   }
 }
 
-// TODO: include Filecoin Deal id
-export const getTaskId = (/** @type {import('./typings').Measurement} */m) => `${m.cid}`
+export const getTaskId = (/** @type {import('./typings').Measurement} */m) => `${m.cid}::${m.minerId}`
```
bajtos, line 154: FYI: I first updated a test to fail with an expected error and only then made this change.

test/helpers/test-data.js
```diff
@@ -1,17 +1,24 @@
 export const VALID_PARTICIPANT_ADDRESS = '0x000000000000000000000000000000000000dEaD'
 
-export const VALID_TASK = {
+export const VALID_TASK_WITH_PROVIDER_PROTOCOL = {
   cid: 'QmUuEoBdjC8D1PfWZCc7JCSK8nj7TV6HbXWDHYHzZHCVGS',
   providerAddress: '/dns4/production-ipfs-peer.pinata.cloud/tcp/3000/ws/p2p/Qma8ddFEQWEU8ijWvdxXm3nxU7oHsRtCykAaVz8WUYhiKn',
   protocol: 'bitswap'
 }
+Object.freeze(VALID_TASK_WITH_PROVIDER_PROTOCOL)
```
bajtos, line 8: We are not using this sample measurement in the old format. Should I completely delete the definition from our source code?

lib/preprocess.js
```diff
@@ -14,6 +14,8 @@ export class Measurement {
     this.participantAddress = pointerize(parseParticipantAddress(m.participant_address))
     this.retrievalResult = pointerize(getRetrievalResult(m))
     this.cid = pointerize(m.cid)
+    this.minerId = pointerize(m.miner_id)
+    this.providerId = pointerize(m.provider_id)
```
juliangruber, line 18: this field seems not to be used anywhere else

test/helpers/test-data.js
juliangruber: 👍

juliangruber (changes requested):

lib/preprocess.js
bajtos: We are not using this field yet, but it's already present in the measurements published by spark-publish - see https://github.com/filecoin-station/spark-api/blob/c92260b573db03bbd67d3e809b8ad15a21a71f74/spark-publish/index.js#L31-L32

If you don't mind, I'd like to add it together with `minerId`.

juliangruber: On this I can share my recent experience extracting `voyager-{api,evaluate}` from spark. There is quite a few fields we're not using yet, and that was confusing to me, as I assumed if they are in the code they are being used. I thought this was a bug and that they were added by mistake.

If you would like to add them now, could you please add a comment, so the codebase stays consistent?

bajtos: see [1ebaa27](https://github.com/filecoin-station/spark-evaluate/pull/177/commits/1ebaa27e1bd09ff09ffdb94428d99f372d3e4494)

bajtos:
@juliangruber I addressed your comments and landed the PR. PTAL and if you have more comments, then I'll open a follow-up PR to address them. 🙏🏻

PR #188, opened 2024-04-12
Feat: Created daily_node_metrics table and received/stored station_id

PatrickNercessian:
DRAFT: DO NOT MERGE

Links:
filecoin-station/roadmap#96


juliangruber (approved):

lib/platform-stats.js
```diff
@@ -0,0 +1,18 @@
...
+ * @param {import('./preprocess').Measurement[]} honestMeasurements
+ */
+export const updateDailyNodeMetrics = async (pgClient, honestMeasurements) => {
+  debug('Updating daily node metrics, count=%s', honestMeasurements.length)
+  for (const m of honestMeasurements) {
+    await pgClient.query(`
+      INSERT INTO daily_node_metrics (station_id, metric_date)
+      VALUES ($1, now()::date)
+      ON CONFLICT (station_id, metric_date) DO NOTHING
+    `, [m.station_id]) // TODO: when we add more fields, we should update the ON CONFLICT clause to update the fields
```
bajtos, line 16: I have a concern about the storage required by this DB schema design. We will store each station ID string in many duplicates (one copy each day).

When I implemented "daily participants," I decided to create a lookup table mapping participant addresses to 32-bit integers and then store those integers in the `daily_participants` table using a foreign-key constraint.

See https://github.com/filecoin-station/spark-evaluate/pull/133

The key insight is that for most queries, we don't care what the exact participant address (or station ID) is. We want to count the number of unique addresses/IDs, and that can be achieved by counting the unique FK references.

The downside is that maintaining such a lookup table efficiently requires complex code (see `mapParticipantsToIds()` in my PR). OTOH, I think it should be fairly easy to generalise that function to support both participant addresses and station IDs by accepting a few more parameters like table name.

WDYT?

test/platform-stats.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+  afterEach(async () => {
+    await pgClient.query('END TRANSACTION')
+  })
+
+  after(async () => {
+    await pgClient.end()
+  })
+
+  it('updates daily node metrics with new measurements', async () => {
```
bajtos, line 37: Since we are anticipating adding more platform stats very soon, let's avoid code churn and nest this new test under `describe('updateDailyNodeMetrics')` (or a similarly named) test group.

```js
describe('updateDailyNodeMetrics', () => {
  it('updates daily node metrics with new measurements', async () => { 
    //...
  })

  it('ignores duplicate measurements for the same station on the same day',
  // ...
  )
})
```

lib/preprocess.js
```diff
@@ -29,6 +29,7 @@ export class Measurement {
     this.end_at = parseDateTime(m.end_at)
     this.status_code = m.status_code
     this.indexerResult = pointerize(m.indexer_result)
+    this.station_id = pointerize(m.station_id)
```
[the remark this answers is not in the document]
PatrickNercessian: Gotcha, used underlines to match the other fields, but makes sense to use camelCase if we're going to switch to it anyway

lib/platform-stats.js
PatrickNercessian: I think this makes sense! As devil's advocate though with some napkin math:

If we're averaging about 2000 station nodes per day, and each row is, let's say, 250B, that's 174MB per year in total. A lookup table for station_id (and maybe deployment_type, CPU architecture, etc.) would save us roughly 60-100 bytes per row? So we'd save 55MB per year.

Even if we 15x the number of daily nodes, we're still looking at <1GB per year of data accumulation in savings. Is it worth the relatively complex code logic?

It's possible I missed some detail here or in the math though.

bajtos: Yeah, I was thinking, too, whether this optimisation is worth the complexity.

We already have 12-20k Station nodes. I think it's realistic to grow the network to 100k nodes by the end of this year. 

IIUC, Ed25519 has 32-byte long public keys, serialised as hex in 64 characters. We need 4 bytes to store the foreign key reference. 64-60=4. If we save 60 bytes per node, that gives us 365 * 60 * 100k bytes = ~2.19GB per year.

I think we can afford to store an extra 2 GB per module per year in our DB.

In the past, when we almost ran out of disk space, it was because our DB needed more than 500 GB. That's a long way to go :)

In that light, I propose to keep your current design.

@juliangruber WDYT?

PatrickNercessian: It should be a constant 64, no? Any reason to accept anything below or above that?

juliangruber: Agreed that the optimization isn't worth it at this point 👍 And we should be able to easily migrate later

> It should be a constant 64, no? Any reason to accept anything below or above that?

+1 to making it the validation as strict as possible

lib/platform-stats.js
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @param {import('pg').Client} pgClient
+ * @param {import('./preprocess').Measurement[]} honestMeasurements
+ */
+export const updateDailyNodeMetrics = async (pgClient, honestMeasurements) => {
+  debug('Updating daily node metrics, count=%s', honestMeasurements.length)
+  for (const m of honestMeasurements) {
+    await pgClient.query(`
+      INSERT INTO daily_node_metrics (station_id, metric_date)
+      VALUES ($1, now()::date)
```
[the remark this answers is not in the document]
juliangruber: > So from the user's perspective, whether one measurement makes it into a certain day or not (if it's on the cusp of a day) is not based on anything they can verify.

Do you know if this is important? We're not talking about payouts here, only public stats. And if it doesn't make it into the current day, it will for sure make it into the next day. Yes this logic could be tighter, but I wouldn't worry about optimizing this now.

lib/public-stats.js
```diff
@@ -31,6 +32,7 @@ export const updatePublicStats = async ({ createPgClient, honestMeasurements })
     }
     await updateDailyParticipants(pgClient, participants)
     await updateIndexerQueryStats(pgClient, honestMeasurements)
+    await updateDailyNodeMetrics(pgClient, honestMeasurements)
```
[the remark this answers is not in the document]
PatrickNercessian: If we don't export the individual functions, it becomes more difficult to unit test them. We have a few options that I see:

1. Call updatePlatformStats in the test class and then check that each individual function worked as expected
2. Conditional exports if it's a test Node env
3. Using a third-party testing framework that supports this
4. Don't unit test them at all?

What do you think?

test/platform-stats.test.js
```diff
@@ -0,0 +1,62 @@
...
+  let pgClient
+
+  before(async () => {
+    pgClient = await createPgClient()
+    await migrateWithPgClient(pgClient)
+  })
+
+  beforeEach(async () => {
+    await pgClient.query('DELETE FROM daily_node_metrics')
+    await pgClient.query('BEGIN TRANSACTION')
```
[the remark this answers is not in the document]
PatrickNercessian: I might as well mimick the date checks that you do in `public-stats.test.js`, since the format of the tables are similar.

lib/platform-stats.js
bajtos: IMO, we should implement the same validation in `zinniad`, to reject invalid Station ID right during the startup. We should also change the hardcoded station id `zinnia` to use a valid string. It doesn't have to be a real public key, but it should pass the validation in spark-evaluate. For example, 64 zero characters (`000(...)000`) or something similar to Ethereum's null address `0x000(...)00dead`. The goal is to use an address that's easy to spot and filter out from production data.

lib/platform-stats.js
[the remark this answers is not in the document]
bajtos: I agree that you should not worry about this right now. 

We use `now()` in other places in public-stats too.

https://github.com/filecoin-station/spark-evaluate/blob/93b4d5ab7d013512870fcf974309f55c04089faf/lib/public-stats.js#L49-L52

BTW, I think PG will convert `now()` from date-time to date automatically, no need to do the conversion explicitly.

```suggestion
      VALUES ($1, now())
```

test/platform-stats.test.js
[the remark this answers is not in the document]
PatrickNercessian: I added similar date checks as public-stats.test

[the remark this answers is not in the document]
bajtos: Sounds good. Please copy the comment from the other test file to explain why we need to wrap the tests in a transaction.

migrations/006.do.daily-node-metrics.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE daily_node_metrics (
+    metric_date DATE NOT NULL,
```
bajtos, line 2: In the other daily table (`daily_participants`), we call the date column `day`.

It makes me wonder why you chose `metric_date` as the name. Did you find the `day` name not clear enough? I don't mind using `metric_date` if there is a good argument. (Although I find it a bit long.)

It would be great to stay consistent, and if we agree to use `metric_date` instead of `day`, then we should look into renaming the existing `day` columns. (That's out of the scope of this pull request, though. It will also require synchronised changes in spark-evaluate and spark-stats.)

```suggestion
    day DATE NOT NULL,
```

bajtos (commented):
Getting close 👍🏻

migrations/006.do.daily-node-metrics.sql
PatrickNercessian: "Day" is often used in Postgres syntax. Although it's not a restricted keyword, typically it's suggested to avoid those as well to prevent confusion.

I don't have a hard opinion on it though, if you'd prefer "day"!

lib/platform-stats.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+export const updateDailyNodeMetrics = async (pgClient, honestMeasurements) => {
+  debug('Updating daily node metrics, count=%s', honestMeasurements.length)
+  for (const m of honestMeasurements) {
+    await pgClient.query(`
+      INSERT INTO daily_node_metrics (station_id, day)
+      VALUES ($1, now())
+      ON CONFLICT (station_id, day) DO NOTHING
+    `, [m.stationId]) // TODO: when we add more fields, we should update the ON CONFLICT clause to update the fields
+  }
```
bajtos, line 25: We have thousands of participants in each round, this code will run thousands of SQL queries and will take very long to complete. I have been bitten by this in the past and had to use a different solution:

https://github.com/filecoin-station/spark-evaluate/blob/562125a2c1bc99b4aaf0fceba4c4bb0a872913b5/lib/public-stats.js#L68-L76

Can you do the same?

lib/platform-stats.js
```diff
@@ -0,0 +1,26 @@
...
+  await updateDailyNodeMetrics(pgClient, honestMeasurements)
+}
+
+/**
+ * @param {import('pg').Client} pgClient
+ * @param {import('./preprocess').Measurement[]} honestMeasurements
+ */
+export const updateDailyNodeMetrics = async (pgClient, honestMeasurements) => {
+  debug('Updating daily node metrics, count=%s', honestMeasurements.length)
+  for (const m of honestMeasurements) {
```
bajtos, line 19: Logging the number of measurements is not very useful. We should log the number of unique Station ids found in the measurements.

It's also important to extract this list of unique IDs before we run SQL queries. Your original implementation will run one SQL query for each measurement, that will perform very poorly.

I would like you to make the following changes:

1. First, build a list of unique station IDs. See how we collect unique participant addresses. https://github.com/filecoin-station/spark-evaluate/blob/562125a2c1bc99b4aaf0fceba4c4bb0a872913b5/lib/public-stats.js#L16-L25
2. Log the number of unique Station IDs found.
3. Use the list of unique Station IDs as the input to SQL queries.

bajtos (changes requested):
Getting close!

lib/platform-stats.js
PatrickNercessian: Committed! I assumed we don't need the extra logic around collecting unique particpant addresses, since the IDs are already provided to use rather than a serial SQL field. Let me know if this works for you!

migrations/006.do.daily-node-metrics.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE daily_node_metrics (
```
bajtos, line 1: Continuing the discussion started in https://github.com/filecoin-station/spark-stats/pull/71#discussion_r1581157857

> Makes sense! Should we also rename the table and references to it in spark-evaluate (it's currently daily_node_metrics)

That's a great idea, let's do it!

Please rename the migration SQL file too.

Other tables are called simply `daily_{things}`, maybe we can do the same here and use the name `daily_stations`?

test/platform-stats.test.js
```diff
@@ -0,0 +1,78 @@
...
+    it('updates daily node metrics for today with multiple measurements', async () => {
+      const validStationId2 = VALID_STATION_ID.slice(0, -1) + '1'
+      const honestMeasurements = [
+        { ...VALID_MEASUREMENT, stationId: VALID_STATION_ID },
+        { ...VALID_MEASUREMENT, stationId: validStationId2 }
+      ]
+
+      await updateDailyNodeMetrics(pgClient, honestMeasurements)
+
+      const { rows } = await pgClient.query('SELECT station_id, day::TEXT FROM daily_node_metrics ORDER BY station_id')
```
bajtos, line 52: This line is very long; please break it into multiple shorter ones. See how we are building multi-line SQL statements in existing code.

bajtos (commented):
Almost there! 👏🏻

lib/preprocess.js
```diff
@@ -200,6 +201,9 @@ const assertValidMeasurement = measurement => {
   assert(ethers.isAddress(measurement.participantAddress), 'valid participant address required')
   assert(typeof measurement.inet_group === 'string', 'valid inet group required')
   assert(typeof measurement.finished_at === 'number', 'field `finished_at` must be set to a number')
+  if (measurement.stationId) {
+    assert(typeof measurement.stationId === 'string' && measurement.stationId.length === 64, 'stationId must be a string of 64 characters')
```
bajtos, line 205: This line is too long, too; let's break it up, please.

lib/platform-stats.js
bajtos: I merged in the latest `main` branch that includes https://github.com/filecoin-station/spark-evaluate/pull/196, we will get the dry-run results soon from the CI.

bajtos: Oh, we need also need to fix our GHA triggers, see https://github.com/filecoin-station/spark-evaluate/pull/198

bajtos (approved):
LGTM 👏🏻

bajtos:
@juliangruber could you please help us to get this pull request across the finish line? The CI build fails in the dry-run step because `GLIF_TOKEN` is not set from the GHA secrets. I clicked on the button approving GHA to execute the CI for this pull request after @PatrickNercessian pushed more changes today.

See https://github.com/filecoin-station/spark-evaluate/actions/runs/8880495010/job/24384374423?pr=188#step:7:14

lib/platform-stats.js
PatrickNercessian: For (1), sure I can definitely work on this in Zinnia!

For (2), this was already done here: https://github.com/filecoin-station/spark-evaluate/blob/f86658b74e921f1b0778194d474dc10d352113de/lib/preprocess.js#L204-L210

Or am I misunderstanding you?

bajtos: > For (2), this was already done

Oh, I missed that bit. All is good! 👍🏻 

> For (1), sure I can definitely work on this in Zinnia!

Great! 💪🏻

PR #194, opened 2024-04-25
feat: dry run for the last round

bajtos:
Improve `bin/dry-run.js` to fetch the index of the last round from on-chain state.

Add dotenv integration to allow developers to keep the Glif API token in the `.env` file.

### DIFF
--- a/bin/dry-run.js
+++ b/bin/dry-run.js
@@ -1,3 +1,6 @@
+// IMPORTANT! We must run dotenv config as the very first thing before importing anything else
+import 'dotenv/config'
+
 import { IE_CONTRACT_ADDRESS, RPC_URL, rpcHeaders } from '../lib/config.js'
 import { evaluate } from '../lib/evaluate.js'
 import { preprocess, fetchMeasurements } from '../lib/preprocess.js'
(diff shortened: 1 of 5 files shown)


bin/dry-run.js
```diff
@@ -1,3 +1,6 @@
+// IMPORTANT! We must run dotenv config as the very first thing before importing anything else
```
juliangruber, line 1:
```suggestion
```

I don't like when me code SHOUTS at me 😅 Would you be ok with either option?

```suggestion
// dotenv has to be imported before anything else
```

lib/config.js
```diff
@@ -14,7 +14,7 @@ const RPC_URL = rpcUrls[Math.floor(Math.random() * rpcUrls.length)]
 console.log(`Selected JSON-RPC endpoint ${RPC_URL}`)
 
 const rpcHeaders = {}
-if (RPC_URL.includes('glif')) {
+if (RPC_URL.includes('glif') && GLIF_TOKEN) {
```
juliangruber, line 17:
```suggestion
if (RPC_URL.includes('glif')) {
```

The glif token is _always_ required

juliangruber (approved):
Good idea to add dotenv, would be cool to add it to other repos as we work on them

juliangruber (changes requested):
Sorry I meant to request changes

bin/dry-run.js
bajtos: I am going to use this:

```js
// dotenv must be imported before importing anything else
```

lib/config.js
bajtos: Unfortunately, our tests don't provide GLIF_TOKEN in the env vars. I had a different implementation initially, it worked when I run `node bin/dry-run.js` but the tests started to fail.

juliangruber: be9f36f lgtm!

> With my change in place, instead of sending invalid authorisation header, we don't send any. I think the outcome should be the same.

I see! I like sending an invalid auth header better, since in the code we're baking in the assumption that it will be provided. Otherwise, as a reader with little context, this line doesn't make sense to me.

PR #200, opened 2024-04-30
Feat: Added honest_measurement_count platform stat

PatrickNercessian:
Link: filecoin-station/roadmap#99

### DIFF
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -16,9 +16,6 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
  */
 export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
   const measurementCountsPerStation = new Map()
-  const stationIds = []
-  const measurementCounts = []
-
   for (const m of honestMeasurements) {
     measurementCountsPerStation.set(
       m.stationId,
@@ -26,12 +23,7 @@ export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
     )
   }
 
-  for (const [stationId, count] of measurementCountsPerStation.entries()) {
-    stationIds.push(stationId)
-    measurementCounts.push(count)
-  }
-
-  debug('Updating daily station stats, unique_count=%s', stationIds.length)
+  debug('Updating daily station stats, unique_count=%s', measurementCountsPerStation.size)
 
   // Coalesce to 0 to avoid null + count = null
   // Use EXCLUDED to allow incrementing the count
@@ -42,5 +34,8 @@ export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
     ON CONFLICT (station_id, day) DO UPDATE
     SET accepted_measurement_count = daily_stations.accepted_measurement_count
                                       + EXCLUDED.accepted_measurement_count
-  `, [stationIds, measurementCounts])
+  `, [
+    Array.from(measurementCountsPerStation.keys()),
+    Array.from(measurementCountsPerStation.values())
+  ])
 }
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -20,6 +20,8 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
 export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
   const measurementCountsPerStation = new Map()
   for (const m of honestMeasurements) {
+    if (m.stationId == null) continue
+
     measurementCountsPerStation.set(
       m.stationId,
       (measurementCountsPerStation.get(m.stationId) || 0) + 1
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -23,9 +23,8 @@ export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
     )
   }
 
-  debug('Updating daily station stats, unique_count=%s', measurementCountsPerStation.size)
+  debug('Updating daily station stats, station_count=%s', measurementCountsPerStation.size)
 
-  // Coalesce to 0 to avoid null + count = null
   // Use EXCLUDED to allow incrementing the count
   // Parameterized stationId to prevent SQL injection
   await pgClient.query(`
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -15,15 +15,28 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
  * @param {import('./preprocess').Measurement[]} honestMeasurements
  */
 export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
-  // TODO: when we add more fields, we will update the ON CONFLICT clause
-  // to update those fields, and we won't just use a Set for the stationIds
-  // which currently removes all granular measurement details
-  const stationIds = [...new Set(honestMeasurements.map(m => m.stationId))]
+  const stationIds = []
+  const counts = []
+  const honestMeasurementCounts = new Map()
+  for (const m of honestMeasurements) {
+    honestMeasurementCounts.set(m.stationId, (honestMeasurementCounts.get(m.stationId) || 0) + 1)
+  }
+
+  for (const [stationId, count] of honestMeasurementCounts.entries()) {
+    stationIds.push(stationId)
+    counts.push(count)
+  }
+
   debug('Updating daily station stats, unique_count=%s', stationIds.length)
 
+  // Coalesce to 0 to avoid null + count = null
+  // Use EXCLUDED to allow incrementing the count
+  // Parameterized stationId to prevent SQL injection
   await pgClient.query(`
-    INSERT INTO daily_stations (station_id, day)
-    SELECT unnest($1::text[]), now()
-    ON CONFLICT (station_id, day) DO NOTHING
-  `, [stationIds])
+    INSERT INTO daily_stations (station_id, day, honest_measurement_count)
+    VALUES (unnest($1::text[]), now(), unnest($2::int[]))
+    ON CONFLICT (station_id, day) DO UPDATE
+    SET honest_measurement_count = COALESCE(daily_stations.honest_measurement_count, 0)
+                                    + EXCLUDED.honest_measurement_count
+  `, [stationIds, counts])
 }
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -29,8 +29,6 @@ export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
   }
   debug('Updating daily station stats, station_count=%s', measurementCountsPerStation.size)
 
-  // Use EXCLUDED to allow incrementing the count
-  // Parameterized stationId to prevent SQL injection
   await pgClient.query(`
     INSERT INTO daily_stations (station_id, day, accepted_measurement_count)
     VALUES (unnest($1::text[]), now(), unnest($2::int[]))
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -15,16 +15,20 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
  * @param {import('./preprocess').Measurement[]} honestMeasurements
  */
 export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
+  const measurementCountsPerStation = new Map()
   const stationIds = []
-  const counts = []
-  const honestMeasurementCounts = new Map()
+  const measurementCounts = []
+
   for (const m of honestMeasurements) {
-    honestMeasurementCounts.set(m.stationId, (honestMeasurementCounts.get(m.stationId) || 0) + 1)
+    measurementCountsPerStation.set(
+      m.stationId,
+      (measurementCountsPerStation.get(m.stationId) || 0) + 1
+    )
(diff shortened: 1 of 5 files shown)


lib/platform-stats.js
```diff
@@ -15,15 +15,28 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
...
-    INSERT INTO daily_stations (station_id, day)
-    SELECT unnest($1::text[]), now()
-    ON CONFLICT (station_id, day) DO NOTHING
-  `, [stationIds])
+    INSERT INTO daily_stations (station_id, day, honest_measurement_count)
+    VALUES (unnest($1::text[]), now(), unnest($2::int[]))
+    ON CONFLICT (station_id, day) DO UPDATE
+    SET honest_measurement_count = COALESCE(daily_stations.honest_measurement_count, 0)
+                                    + EXCLUDED.honest_measurement_count
+  `, [stationIds, counts])
```
juliangruber, line 41:
```suggestion
  `, [
    Array.from(honestMeasurementCounts.keys()),
    Array.from(honestMeasurementCounts.values())
  ])
```

lib/platform-stats.js
```diff
@@ -15,15 +15,28 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
  * @param {import('./preprocess').Measurement[]} honestMeasurements
  */
 export const updateDailyStationStats = async (pgClient, honestMeasurements) => {
-  // TODO: when we add more fields, we will update the ON CONFLICT clause
-  // to update those fields, and we won't just use a Set for the stationIds
-  // which currently removes all granular measurement details
-  const stationIds = [...new Set(honestMeasurements.map(m => m.stationId))]
+  const stationIds = []
+  const counts = []
```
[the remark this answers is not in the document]
juliangruber: can just use one data structure (see following suggestions)

lib/platform-stats.js
bajtos: I find it a bit difficult to understand what exactly is stored in the data structure called "honestMeasurementCounts".
- IIUC, we store counts per station instance; let's encode that in the name.
- I don't think it's important to mention that we are counting honest measurements only because this function does not have access to any other measurements.

A few proposed names for inspiration:
- `measurementCountsPerStation`
- `perStationStats` (assuming we will be adding more per-station fields in the future)
- `perStationMeasurementCounts`
- `stationMeasurementCounts`

PatrickNercessian: SGTM. Are we okay with the column name being honest_measurement_count still?

lib/platform-stats.js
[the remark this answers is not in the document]
PatrickNercessian: This would certainly be a more readable way, and what I started with originally, but it would be less efficient as we'd have to loop through the number of station IDs twice.

If we don't care about that since it's already a smaller, aggregated data structure, happy to make this change!

migrations/007.do.add-daily-job-stats.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE daily_stations
+ADD COLUMN honest_measurement_count INT
```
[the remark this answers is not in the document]
PatrickNercessian: Works for me. The downside is that we'll have a bunch of old entries that say 0 counts in the table, which isn't technically correct

lib/platform-stats.js
bajtos: Good point! Let me provide more context.

When spark-evaluate (and voyager-evaluate) downloads the measurements committed to the smart contract, we get a list of all measurements submitted for the given round. We don't know yet which measurements were performed by honest nodes and which were fraudulent.

One of the steps in the evaluation pipeline is to run various fraud-detection algorithms to detect suspicious measurements that should not be rewarded.

The outcome is a list of all measurements we consider honest and which should be rewarded.

Now, in our private dashboard, we often want to look at statistics about both sets for measurements (all measurements, honest measurements) to spot anomalies that could indicate a bug in our fraud detection.

When we started the work on the public dashboard, we realised that we must always use only honest measurements to calculate the public stats because that's the only dataset we trust enough.

However, when reading the spark-evaluate code, I think it's useful to be reminded that the input to these calculations is the set of honest measurements, that's why I used the name `honestMeasurements` instead of simply `measurements`.

When it comes to the database schema, we don't need to keep this distinction IMO, as the presentation layer (API & Dashboards) does not differentiate between "all" vs "honest" measurements.

So let's use `measurement_count` or even `measurements` as the SQL column name.

migrations/007.do.add-daily-job-stats.sql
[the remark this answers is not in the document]
bajtos: Yeah, I agree it's not technically correct. At the same time, this applies only to historical data, right? And we can backfill these values if needed.

Let me ask a different question. If we keep NULL values in this column, how do you expect consumers of this data to handle the NULL value?
- Grafana charts
- 3rd-party API consumers

test/platform-stats.test.js
```diff
@@ -41,36 +43,42 @@ describe('platform-stats', () => {
...
+        { ...VALID_MEASUREMENT, stationId: VALID_STATION_ID_2 }
       ]
 
       await updateDailyStationStats(pgClient, honestMeasurements)
 
-      const { rows } = await pgClient.query('SELECT station_id, day::TEXT FROM daily_stations')
-      assert.strictEqual(rows.length, 1)
-      assert.deepStrictEqual(rows, [{ station_id: VALID_STATION_ID, day: today }])
+      const { rows } = await pgClient.query(`
+        SELECT station_id, day::TEXT, honest_measurement_count FROM daily_stations
```
bajtos, line 74: Nitpick, feel free to ignore.

```suggestion
        SELECT station_id, day::TEXT, honest_measurement_count 
        FROM daily_stations
```

lib/platform-stats.js
PatrickNercessian: Understood! Are we certain we'll never want to show a distinction between "all" vs "honest" measurements? I think a while ago we had briefly discussed doing this.

migrations/007.do.add-daily-job-stats.sql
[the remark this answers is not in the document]
PatrickNercessian: Yes, specifically, this applies to "historical data" in the context of "after we released the station_id addition but before we release this PR"

I believe in Grafana you can just use a data transformation for NULL values to treat them as 0. For 3rd party API consumers, we would let them handle it.

I don't have a hard opinion either way though, just devil's advocate here

lib/platform-stats.js
bajtos: > Understood! Are we certain we'll never want to show a distinction between "all" vs "honest" measurements? I think a while ago we had briefly discussed doing this.

I can't rule that out for 100%. OTOH, all other tables and columns don't encode "honest" in their names.

Oh, I think I see what you mean. Here, we are specifically talking about the number of measurements, so it makes sense to be more explicit about "honest" vs "all" measurements.

OK, let's keep "honest" in the SQL column name, but change it to "accepted". I feel it better describes the intent and does not have any negative connotations of accusing the participant of doing something wrong.

migrations/007.do.add-daily-job-stats.sql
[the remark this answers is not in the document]
bajtos: I see your point, and I do value accuracy in our data and DB schema design.

However, in this case, I prefer to prioritise the ease of use for data consumers because the NULL-vs-0 distinction is an edge case in historical data, and nobody will care about that fine detail very soon (once we start recording real data).

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: To me optimization unless you've measured it and discovered a notable impact on performance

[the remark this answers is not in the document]
juliangruber: If no-one notices your code being more efficient, then to me, it only has become less maintainable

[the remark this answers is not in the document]
PatrickNercessian: Fair point! Addressed in https://github.com/filecoin-station/spark-evaluate/pull/200/commits/afda88ed38927fcec00cc463f02538d86676f33e

lib/platform-stats.js
```diff
@@ -15,15 +15,27 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
...
+    measurementCountsPerStation.set(
+      m.stationId,
+      (measurementCountsPerStation.get(m.stationId) || 0) + 1
+    )
+  }
 
+  debug('Updating daily station stats, unique_count=%s', measurementCountsPerStation.size)
+
+  // Coalesce to 0 to avoid null + count = null
+  // Use EXCLUDED to allow incrementing the count
```
bajtos, line 29: Do we still need to use EXCLUDED now that `accepted_measurement_count` is never NULL?

PatrickNercessian: I believe so. EXCLUDED is for the fact that there was a conflict. For example, see public-stats.js `mapParticipantsToIds`

I did remove COALESCE because it's never NULL now, but I forgot to remove the comment.

lib/platform-stats.js
```diff
@@ -15,15 +15,27 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
...
-  debug('Updating daily station stats, unique_count=%s', stationIds.length)
+  const measurementCountsPerStation = new Map()
+  for (const m of honestMeasurements) {
+    measurementCountsPerStation.set(
+      m.stationId,
+      (measurementCountsPerStation.get(m.stationId) || 0) + 1
+    )
+  }
 
+  debug('Updating daily station stats, unique_count=%s', measurementCountsPerStation.size)
```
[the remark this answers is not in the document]
PatrickNercessian: Good call, it is confusing now that we have a measurement count as well. Addressed in https://github.com/filecoin-station/spark-evaluate/pull/200/commits/06f9740c4d65b46590631cfa401368b413e8493c

bajtos (approved):

test/platform-stats.test.js
```diff
@@ -50,36 +52,44 @@ describe('platform-stats', () => {
...
+        ORDER BY station_id`
+      )
+      assert.strictEqual(rows.length, 2)
+      assert.deepStrictEqual(rows, [
+        { station_id: VALID_STATION_ID, day: today, accepted_measurement_count: 2 },
+        { station_id: VALID_STATION_ID_2, day: today, accepted_measurement_count: 1 }
+      ])
     })
 
     it('ignores measurements without .stationId', async () => {
```
bajtos, line 95: @PatrickNercessian this test is failing now. Could you please investigate?

```
  1) platform-stats
       updateDailyStationStats
         ignores measurements without .stationId:
     error: null value in column "station_id" of relation "daily_stations" violates not-null constraint
      at /Users/bajtos/src/spark/evaluate/node_modules/pg/lib/client.js:526:17
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async updateDailyStationStats (file:///Users/bajtos/src/spark/evaluate/lib/platform-stats.js:32:3)
      at async Context.<anonymous> (file:///Users/bajtos/src/spark/evaluate/test/platform-stats.test.js:101:7)
```

bajtos: I merged the latest version from `main` into your feature branch and resolved the merge conflict, hopefully correctly. Please remember to `git pull` the updates!

PatrickNercessian: Ah apologies, will do.

Addressed in https://github.com/filecoin-station/spark-evaluate/pull/200/commits/1c9884ec6a6e9c12af294cd883f3a1de2c1805a8

bajtos:
@juliangruber PTAL

lib/platform-stats.js
```diff
@@ -18,23 +18,29 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
...
+      m.stationId,
+      (measurementCountsPerStation.get(m.stationId) || 0) + 1
     )
-  ]
-  debug('Updating daily station stats, unique_count=%s', stationIds.length)
+  }
+  debug('Updating daily station stats, station_count=%s', measurementCountsPerStation.size)
 
+  // Use EXCLUDED to allow incrementing the count
+  // Parameterized stationId to prevent SQL injection
```
juliangruber, line 33:
```suggestion
```

The comment made sense in the context of this PR, but to other readers this should be the default way of building a query anyway

lib/platform-stats.js
```diff
@@ -18,23 +18,29 @@ export const updatePlatformStats = async (pgClient, honestMeasurements) => {
...
+    measurementCountsPerStation.set(
+      m.stationId,
+      (measurementCountsPerStation.get(m.stationId) || 0) + 1
     )
-  ]
-  debug('Updating daily station stats, unique_count=%s', stationIds.length)
+  }
+  debug('Updating daily station stats, station_count=%s', measurementCountsPerStation.size)
 
+  // Use EXCLUDED to allow incrementing the count
```
juliangruber, line 32:
```suggestion
```

This is kind of how these queries work, it's a common approach. I don't think it warrants a comment.

juliangruber (changes requested):

juliangruber (approved):

PR #201, opened 2024-04-30
Feat: Add new table daily_fil and populate with new Transfer events

PatrickNercessian:
Link: https://github.com/filecoin-station/roadmap/issues/97

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,6 +4,7 @@ import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { onContractEvent } from 'on-contract-event'
 import { RoundData } from './lib/round.js'
+import { updateDailyFilStats } from './lib/platform-stats.js'
 
 export const startEvaluate = async ({
   ieContract,
@@ -103,6 +104,13 @@ export const startEvaluate = async ({
     })
   }
 
+  const onTransfer = async (to, amount) => {
+    const pgClient = await createPgClient()
+    const transferEvent = { to, amount }
+    console.log('Event: Transfer', transferEvent)
+    updateDailyFilStats(pgClient, transferEvent)
+  }
+
   // Listen for events
   const it = onContractEvent({
     contract: ieContract,
@@ -115,6 +123,8 @@ export const startEvaluate = async ({
       onMeasurementsAdded(...event.args)
     } else if (event.name === 'RoundStart') {
       onRoundStart(...event.args)
+    } else if (event.name === 'Transfer') {
+      await onTransfer(...event.args)
     }
   }
 }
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -103,6 +104,13 @@ export const startEvaluate = async ({
     })
   }
 
+  const onTransfer = async (to, amount) => {
```
PatrickNercessian, line 107: Are we okay with making this async, so that we can createPgClient?

Alternatively, instead of passing around the createPgClient here and to public-stats.js, we could just create pgClient in index.js or somewhere else early in the execution, and then just pass pgClient around. One downside here is that we would be initalizing the pgClient even if we don't need to use it, i.e. in onMeasurementsAdded()

migrations/008.do.daily-fil-metrics.sql
PatrickNercessian, line 1: Should the amount be BIGINT? It has 8 bytes, but in our Impact Evaluator Balances.sol, "amount" is uint256 (32 bytes unsigned).

Should we instead use something like "amount NUMERIC(78,0)"? Or do we not expect values that high?

PatrickNercessian:
Do we want to create dry run functionality for this?

We could incorporate into the existing bin/dry-run.js file, or create a new one for this functioanlity exclusively. Alternatively, we can prioritize deployment speed here, and just verify that no existing functionality would break if this doesn't work.

migrations/008.do.daily-fil-metrics.sql
PatrickNercessian: Also this is named 008 in order to avoid conflict with the 007 in #200

juliangruber (changes requested):
Could we move all of this logic to spark-stats? It can also listen for contract events, and this logic doesn't have to be in spark-evaluate, since it doesn't depend on evaluation results, and therefore is out of scope.

PatrickNercessian:
> Could we move all of this logic to spark-stats? It can also listen for contract events, and this logic doesn't have to be in spark-evaluate, since it doesn't depend on evaluation results, and therefore is out of scope.

The logic as-is requires an ethers contract for the impact evaluator contract. That uses the lib/abi.json file, among other env variables. Are we comfortable with copy-pasting that file to the spark-stats repo as well? Or would you prefer an alternate solution, like a shared file location for the abi.json file, or something else? @juliangruber

juliangruber:
> > Could we move all of this logic to spark-stats? It can also listen for contract events, and this logic doesn't have to be in spark-evaluate, since it doesn't depend on evaluation results, and therefore is out of scope.
> 
> The logic as-is requires an ethers contract for the impact evaluator contract. That uses the lib/abi.json file, among other env variables. Are we comfortable with copy-pasting that file to the spark-stats repo as well? Or would you prefer an alternate solution, like a shared file location for the abi.json file, or something else? @juliangruber

We have been sharing the contract information like this in all the repos. Until we have a better method (likely with a new node module), this is the way to go 👍

PR #202, opened 2024-05-02
platform-stats: fix ignore empty station id

juliangruber:
Fixes dry run error

```
error: null value in column "station_id" of relation "daily_stations" violates not-null constraint
```


bajtos (commented):
Can you please add a unit test to verify the fix and prevent regressions in the future?

bajtos (approved):
:shipit:

PR #204, opened 2024-05-02
ci: move dry-run to a parallel workflow

bajtos:
Also skip the entire workflow when a pull request comes from a fork (and therefore does not have access to our GHA secrets).


.github/workflows/dry-run.yml
```diff
@@ -0,0 +1,38 @@
+name: Dry Run
+on:
+  push:
+  # Run on all branches in our repo, so that we test dry-run for our PRs too
+  # branches: [main]
+  # Don't run on forks, which don't have access to secrets
+  # pull_request:
+  #  branches: [main]
```
juliangruber, line 8: is this part still wip?

bajtos: It isn't, we can remove the lines that are commented-out.

bajtos: I removed the commented-out code in [a1472bd](https://github.com/filecoin-station/spark-evaluate/pull/204/commits/a1472bd301032f3995ce1e844f34b5e214658d57).

LGTY now?

juliangruber (approved):
fantastic 👏

juliangruber (changes requested):
I think this is missing migrations:

```
error: relation "retrieval_stats" does not exist
```

bajtos:
> I think this is missing migrations:
> 
> ```
> error: relation "retrieval_stats" does not exist
> ```

Great catch, I added an extra step - `npm run migrate`.

How do you feel about having this dry-run defined in a new file? I think it's a part of CI, so it would be better to have it alongside other CI-related blocks like "build" and "docker-build".

The difference is in how GitHub visualises this in the UI. At the moment, dry-run progress & outcome is in a different place than the rest of the CI blocks.

Screenshot 2024-05-03 at 10 27 11

Screenshot 2024-05-03 at 10 28 00

I am leaning towards moving the dry-run block to `ci.yaml` and figuring out the right `if:` condition that will run this block only for pull requests that have the GLIF_TOKEN secret configured.

juliangruber:
+1 to moving it to the same file, for grouping purposes 👍

.github/workflows/ci.yml
```diff
@@ -35,15 +35,40 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - run: npm ci
+      - run: npm run migrate
       - name: Evaluate the last Spark round in dry-run mode
-        # Don't run on forks, which don't have access to secrets
-        if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
```
bajtos, line 40: I moved this condition to the top of this block so that the entire `dry-run` block is skipped when the condition is false.

bajtos:
> +1 to moving it to the same file, for grouping purposes 👍

Done

Screenshot 2024-05-03 at 15 57 29

juliangruber (approved):
👏

PR #205, opened 2024-05-02
Update dry-run error handling

juliangruber:


bin/dry-run.js
```diff
@@ -13,6 +14,13 @@ import { ethers } from 'ethers'
 import pg from 'pg'
 import { RoundData } from '../lib/round.js'
 
+Sentry.init({
+  dsn: 'https://d0651617f9690c7e9421ab9c949d67a4@o1408530.ingest.sentry.io/4505906069766144',
+  environment: 'dry-run',
```
bajtos, line 19: I propose to set the environment to "CI" when running on GitHub Actions:

```suggestion
  environment: process.env.SENTRY_ENVIRONMENT || 'dry-run',
```

See also:
https://github.com/filecoin-station/spark-evaluate/blob/b1280f727f3e5bd29c7257fdbdeba7c6eedf077e/bin/spark-evaluate.js#L15-L18

index.js
```diff
@@ -51,13 +51,6 @@ export const startEvaluate = async ({
       recordTelemetry,
       logger
     }).catch(err => {
-      // See https://github.com/filecoin-station/spark-evaluate/issues/36
-      // Because each error message contains unique CID, Sentry is not able to group these errors
-      // Let's wrap the error message in a new Error object as a cause
-      if (typeof err === 'string' && err.match(/ENOENT: no such file or directory, open.*\/bafy/)) {
-        err = new Error('web3.storage cannot find block\'s temp file', { cause: err })
-      }
-
```
bajtos, line 60: IIRC, we were discussing moving this block from `index.js` down into `preprocess()` function, but maybe I misunderstood your intention.

I don't mind to remove this block entirely if you are sure such change will not degrade the quality of Sentry issue / will not introduce duplicate issues.

juliangruber: Ah sorry I mentioned this in the commit message but should have made it more explicit here. This error isn't happening any more (not found in papertrail or sentry), so to me it wasn't worth the complexity of refactoring it

bajtos: That makes sense. I agree that we can remove this block if we are no longer triggering this error. I vaguely remember that we might have eliminated the problem by refactoring our fetch function to call the IPFS GW instead of using the web3 storage client SDK.

Please add the explanation to the commit message, our future selves will thank you.

bajtos (approved):

Issue #64, opened 2023-11-21
Fraud assessment "undefined"


bajtos:
When we evaluate measurements, the fraud assessment pipeline often assigns an `undefined` value to some.

bajtos:
Data for round 528:

```
dry-run
  measurements_OK: 187,880
  measurements_INVALID_TASK: 9,446
  measurements_DUP_INET_GROUP: 165,958
total: 363,284

production:
  measurements_OK: 182,212
  measurements_INVALID_TASK: 9,446
  measurements_DUP_INET_GROUP: 153,823
total: 345,482
```

juliangruber:
Do you have a suspicion where `_undefined` is coming from? Shall we pair on this?

bajtos:
My current thinking: There is a race condition - we receive new measurements and round-start events close together. The round evaluation starts running before we fetch & preprocess the last batch of measurements. However, these measurements are still added to our in-memory store. So later, when we collect stats, we iterate over these measurements too. I have an idea how to fix this. I hope to have a PR ready tomorrow. 🤞🏻

bajtos:
We haven't seen "fraud assessment undefined" in the last 90 months. I believe the partial fixes we made as part of https://github.com/filecoin-station/spark-evaluate/issues/233, most notably https://github.com/filecoin-station/spark-evaluate/pull/224, solved the problem.

PR #209, opened 2024-05-03
preprocess: add delay to improve changes of data being available

juliangruber:
As suggested in https://filecoinproject.slack.com/archives/C031K3BG527/p1714658724173009?thread_ts=1714636341.173339&cid=C031K3BG527. Data stored in W3S isn't always immediately available, and especially sometimes trying to fetch it too soon will result in bad responses in the caches.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,7 +7,7 @@ import { RoundData } from './lib/round.js'
 import timers from 'node:timers/promises'
 
 // Tweak this value to improve the chances of the data being available
-const PREPROCESS_DELAY = 60_1000
+const PREPROCESS_DELAY = 60_000
 
 export const startEvaluate = async ({
   ieContract,
--- a/index.js
+++ b/index.js
@@ -4,6 +4,10 @@ import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { onContractEvent } from 'on-contract-event'
 import { RoundData } from './lib/round.js'
+import timers from 'node:timers/promises'
+
+// Tweak this value to improve the chances of the data being available
+const PREPROCESS_DELAY = 60_1000
 
 export const startEvaluate = async ({
   ieContract,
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -4,6 +4,10 @@ import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { onContractEvent } from 'on-contract-event'
 import { RoundData } from './lib/round.js'
+import timers from 'node:timers/promises'
+
+// Tweak this value to improve the chances of the data being available
+const PREPROCESS_DELAY = 60_1000
```
bajtos, line 10: This value is 601 seconds, did you mean `60_000` instead?

60 seconds seems a bit long delay to me, how about starting with a smaller value like `10_000` or `30_000`?

bajtos (changes requested):
+1 to give this a try.

There is a big catch though: by introducing this delay, when we receive the last batch of measurements for a round, we are almost certainly going to evaluate this round before we fetch the measurements for this round, effectively discarding that last batch.

The current implementation does not handle this situation well either, see https://github.com/filecoin-station/spark-evaluate/issues/64#issuecomment-1830299843

WDYT, is this the right time to bite the bullet and fix the underlying race condition common to both the old and the proposed version? Or should we defer that fix for later, to ship your fix faster?

index.js
juliangruber: Actually 60s is the lower boundary, it was suggested to go with 1-5 minutes

juliangruber: Fixed the number 😅

juliangruber:
> There is a big catch though: by introducing this delay, when we receive the last batch of measurements for a round, we are almost certainly going to evaluate this round before we fetch the measurements for this round, effectively discarding that last batch.

This is an unfortunate truth about this logic: It's better to pretty sure miss the last measurements batch per round, than to risk missing every batch.

> WDYT, is this the right time to bite the bullet and fix the underlying race condition common to both the old and the proposed version? Or should we defer that fix for later, to ship your fix faster?

Good question. I think it's not the right time, we have more important issues (like increasing operator rewards). If nodes stay online, it doesn't matter so much if one batch is missed, as the distribution of rewards should still be roughly the same. We should 100% get this tight eventually though.

bajtos (approved):
👍🏻

PR #223, opened 2024-05-20
feat: record histogram of "score per inet group"

bajtos:
Give us visibility into how much rewards are paid to which IPv4 subnet.


bajtos:

bajtos:
https://protocollabs.grafana.net/d/c4bd3c80-9360-4ea3-8b75-c5ecea90765b/spark?orgId=1&refresh=10s&from=now-1h&to=now

Screenshot 2024-05-20 at 17 37 35

csmasterpath2023:
I have a question that why different IPv4 subnet gets different score, If only the first 15 tasks of each IPv4 address are rewarded, shouldn't different addresses receive the same score?

juliangruber (commented):
👍

bajtos:
> I have a question that why different IPv4 subnet gets different score, If only the first 15 tasks of each IPv4 address are rewarded, shouldn't different addresses receive the same score?

We had a bug in our algorithm. We rewarded up to 15 measurements for each triple `(inet_group, cid, miner_id)`. As a result, if you ran two Station instances on a single machine, and each instance executed different tasks, you would get rewards for 30 tasks. The score for your inet group (your IPv4 /24 subnet) would be 30.

I discovered this problem yesterday, I am working on the fix - see #225.

csmasterpath2023:
> > I have a question that why different IPv4 subnet gets different score, If only the first 15 tasks of each IPv4 address are rewarded, shouldn't different addresses receive the same score?
> 
> We had a bug in our algorithm. We rewarded up to 15 measurements for each triple `(inet_group, cid, miner_id)`. As a result, if you ran two Station instances on a single machine, and each instance executed different tasks, you would get rewards for 30 tasks. The score for your inet group (your IPv4 /24 subnet) would be 30.
> 
> I discovered this problem yesterday, I am working on the fix - see #225.
What if my computer has multiple IPs

bajtos:
> What if my computer has multiple IPs

In Spark, we value measurements performed from a variety of network locations (IPv4 /24 subnets). If your computer is connected to the internet via multiple addresses in different IPv4 /24 subnets, then you can earn more rewards. I would recommend you to set up your operations in such way that each Station instance is using one IPv4 address (route to the internet) all the time.

csmasterpath2023:
But my computer has multiple public IP addresses, and I cannot guarantee that every communication will use the same IP address.Will this earn more rewards?

bajtos:
> But my computer has multiple public IP addresses, and I cannot guarantee that every communication will use the same IP address.Will this earn more rewards?

First of all, it's not enough to have different IP addresses. You need this IP addresses to be in different IPv4 /24 subnets. For example, we group `192.168.1.10` and `192.168.1.20` together and split the rewards. On the other hand, `192.168.1.10` and `192.168.2.10` will receive full rewards each (assuming there is no other Station with an IP address  `192.168.1.*` or `192.168.2.*`).

csmasterpath2023:
Yes, my computer has multiple IPv4 public network addresses, and I have run multiple station cores. Will this earn more rewards?Assuming I have 100 IPv4 public network addresses and each address successfully participates in 15 tasks, will I receive 100 rewards?

PR #225, opened 2024-05-20
fix: rewards per inet group

bajtos:
Rework the rewards awarded to each inet-group so that:
- We reward at most `maxTasksPerNode` measurements in each inet group, regardless of which task they belong to
- We reward only the first measurement for each task 

Links:
- https://github.com/filecoin-station/spark-evaluate/pull/223
- https://github.com/filecoin-station/spark-evaluate/pull/226

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -1,7 +1,7 @@
 import createDebug from 'debug'
 import assert from 'node:assert'
 import * as Sentry from '@sentry/node'
-import { buildRetrievalStats, recordCommitteeSizes } from './retrieval-stats.js'
+import { buildRetrievalStats, getTaskId, recordCommitteeSizes } from './retrieval-stats.js'
 import { updatePublicStats } from './public-stats.js'
 
 const debug = createDebug('spark:evaluate')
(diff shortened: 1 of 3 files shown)


lib/evaluate.js
```diff
@@ -294,35 +296,32 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
     // 2. Pick the measurement with the lowest hash value
     // This relies on the fact that the hash function has a random distribution.
     // We are also making the assumption that each measurement has a distinct `finished_at` field.
+    //
+    const measurementsWithHash = await Promise.all(
+      groupMeasurements.map(async (m) => ({ m, h: await getHash(m) }))
+    )
+    // Sort measurements using the computed hash
+    measurementsWithHash.sort((a, b) => a.h < b.h ? -1 : a.h > b.h ? 1 : 0)
```
juliangruber, line 304:
```suggestion
    measurementsWithHash.sort((a, b) => a.h.localeCompare(b.h))
```

juliangruber (approved):
👏

lib/evaluate.js
bajtos: I am a bit concerned about what locale will be used for comparison. I think it's fine to assume the Fly.io environment has a reasonable locale, but how about local development? I don't want the tests to fail if my locale is set to CZ or DE.

Fortunately, we can hard-code the locale in the second argument.

```js
a.h.localeCompare(b.h, 'en')
```

bajtos: Addressed in [668ebf2](https://github.com/filecoin-station/spark-evaluate/pull/225/commits/668ebf2a2ba02414da795de84b93650631a32ff3)

PR #226, opened 2024-05-21
feat: remove "inet group success rate" stats

bajtos:
These stats are difficult to understand. Thefeore, we haven't been using them for a while now.

My work on fixing the per-inet-group reward selection in #225 would require us to significantly rework the way how to calculate these stats. I decided it's not worth the effort.

As of #223, we are recording slightly different stats about per-inet-group rewards, so we will still have some visibility into this area.

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -69,7 +69,7 @@ export const evaluate = async ({
 
   const started = Date.now()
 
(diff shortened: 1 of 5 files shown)


bajtos:

juliangruber (commented):
👍

PR #237, opened 2024-05-24
feat: bin/fetch-recent-miner-measurements.js

bajtos:
Add a script enabling storage providers to inspect recent Spark measurements for content they are storing.

**Example usage:**

```
node bin/fetch-recent-miner-measurements.js f010479
```

**Example output in measurements-f010479.txt**

```
Timestamp                CID                                                                    Protocol   RetrievalResult
2024-05-24T02:43:14.371Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:43:19.564Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:43:28.086Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:43:30.087Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:43:33.123Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:43:39.298Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki            http       BAD_GATEWAY
2024-05-24T02:44:29.340Z bafybeigpefr2xxsiigy5witxbmgrg3ok7tnrauz44z3utch5uxakljxlki                       IPNI_ERROR_404
```

**TODO**
- [x] process batches in a streaming way
- [x] lightweight documentation

**Out of scope**
- Dockerfile + a GH Actions step to build & publish docker images

Links:
- https://github.com/filecoin-station/roadmap/issues/107

### DIFF
--- a/bin/fetch-recent-miner-measurements.js
+++ b/bin/fetch-recent-miner-measurements.js
@@ -4,7 +4,8 @@ import 'dotenv/config'
 import { Point } from '@influxdata/influxdb-client'
 import * as Sentry from '@sentry/node'
 import createDebug from 'debug'
-import fs, { mkdir, readFile, writeFile } from 'node:fs/promises'
+import fs from 'node:fs'
+import { mkdir, readFile, writeFile } from 'node:fs/promises'
 import os from 'node:os'
 import path from 'node:path'
 import pMap from 'p-map'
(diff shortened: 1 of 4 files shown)


lib/preprocess.js
```diff
@@ -166,11 +174,13 @@ export const fetchMeasurements = async cid => {
     }
   })
   for await (const entry of entries) {
+    signal?.throwIfAborted()
```
juliangruber, line 177: nice! I didn't know about this method

juliangruber: Shall we also add this to the `async get` function above, to cover more ground?

bin/fetch-recent-miner-measurements.js
```diff
@@ -0,0 +1,214 @@
...
+import os from 'node:os'
+import path from 'node:path'
+import pMap from 'p-map'
+import { IE_CONTRACT_ADDRESS } from '../lib/config.js'
+import { createMeridianContract } from '../lib/ie-contract.js'
+import { fetchMeasurements, preprocess } from '../lib/preprocess.js'
+import { RoundData } from '../lib/round.js'
+
+const {
+  STORE_ALL = false
```
juliangruber, line 18:
```suggestion
  STORE_ALL = 'false'
```

Only strings can be passed as env vars, I think the boolean conversion should happen explicitly

bin/fetch-recent-miner-measurements.js
```diff
@@ -0,0 +1,214 @@
...
+Usage:
+  ${nodePath} ${selfPath} [contract-address] minerId [blocksToQuery]
+`
+
+if (!contractAddress) {
+  console.error('Missing required argument: contractAddress')
+  console.error(USAGE)
+  process.exit(1)
+}
+
```
juliangruber, line 49:
```suggestion
```

With the logic above I believe this is dead code

bin/fetch-recent-miner-measurements.js
```diff
@@ -0,0 +1,214 @@
...
+console.error('Querying the chain for recent MeasurementsAdded events')
+const measurementCids = await getRecentMeasurementsAddedEvents(contractAddress, blocksToQuery)
+console.error(' → found %s events', measurementCids.length)
+
+const ALL_MEASUREMENTS_FILE = 'measurements-all.ndjson'
+const MINER_DATA_FILE = `measurements-${minerId}.ndjson`
+const MINER_SUMMARY_FILE = `measurements-${minerId}.txt`
+
+const allMeasurementsWriter = STORE_ALL
```
juliangruber, line 68:
```suggestion
const allMeasurementsWriter = STORE_ALL === 'true'
```

bin/fetch-recent-miner-measurements.js
```diff
@@ -0,0 +1,214 @@
...
+  ? fs.createWriteStream(ALL_MEASUREMENTS_FILE)
+  : undefined
+
+const minerDataWriter = fs.createWriteStream(MINER_DATA_FILE)
+const minerSummaryWriter = fs.createWriteStream(MINER_SUMMARY_FILE)
+minerSummaryWriter.write(formatHeader() + '\n')
+
+const abortController = new AbortController()
+const signal = abortController.signal
+process.on('SIGINT', () => abortController.abort(new Error('interrupted')))
```
juliangruber, line 78:
```suggestion
```

Doesn't it shutdown without this as well?

bin/fetch-recent-miner-measurements.js
```diff
@@ -0,0 +1,214 @@
...
+
+try {
+  await pMap(measurementCids, fetchAndProcess, { concurrency: os.cpus().length })
```
juliangruber, line 81:
```suggestion
  await pMap(measurementCids, fetchAndProcess, { concurrency: os.availableParallelism() })
```

> os.cpus().length should not be used to calculate the amount of parallelism available to an application. Use [os.availableParallelism()](https://nodejs.org/api/os.html#osavailableparallelism) for this purpose.

https://nodejs.org/api/os.html#oscpus

juliangruber (changes requested):

bin/fetch-recent-miner-measurements.js
bajtos: You can still provide an empty string in the CLI arg, right?

bin/fetch-recent-miner-measurements.js
bajtos: It will shutdown, but we won't have a chance to print the filenames where we wrote the data.

bin/fetch-recent-miner-measurements.js
juliangruber: You're right 👍 

```
$ node -pe "console.log(process.argv)" ""
[ '/Users/julian/.nvm/versions/node/v20.12.1/bin/node', '' ]
```

bajtos:
> lightweight documentation

https://www.notion.so/spacemeridian/Troubleshooting-Miner-Score-664ccb2e5c264b39986df09db6b445a4

bin/fetch-recent-miner-measurements.js
bajtos: Well, we detect the empty string and insert the contract address. Then we interpret the empty string as a miner id.

```
❯ node bin/fetch-recent-miner-measurements.js '' f010479 
Selected JSON-RPC endpoint https://api.node.glif.io/rpc/v0
Missing required argument: minerId

Usage:
  /usr/local/bin/node /Users/bajtos/src/spark/evaluate/bin/fetch-recent-miner-measurements.js [contract-address] minerId [blocksToQuery]
```

I think this is an edge case not worth spending our time on, but let me know if you disagree.

juliangruber: I understand, let's get back to this when it becomes a problem?

bin/fetch-recent-miner-measurements.js
bajtos: IIRC, the convention is to treat non-zero numbers as true. I suppose we should also support upper-case `TRUE` values too. It gets complex pretty quickly.

I am going to implement the following condition:
```js
!!STORE_ALL && STORE_ALL.toLowerCase() !== 'false' && STORE_ALL !== '0'
```

Test cases:

```
> function check(STORE_ALL) { return !!STORE_ALL && STORE_ALL.toLowerCase() !== 'false' && STORE_ALL !== '0' }
undefined
> check('0')
false
> check('1')
true
> check()
false
> check('true')
true
> check('false')
false
> check('foobar')
true
> check('')
false
> check('FaLSE')
false
> check('TRUe')
true
```

bin/fetch-recent-miner-measurements.js
bajtos: Thank you for educating me about these two APIs, I learned something new 😄

bin/fetch-recent-miner-measurements.js
bajtos: On the second thought, let's use `undefined` as the default value.

juliangruber (approved):
NICE!

PR #249, opened 2024-06-05
preprocess: fix disable fetchMeasurements cache on retries

juliangruber:
As suggested [here](https://filecoinproject.slack.com/archives/C027XP5BTGB/p1717581326121119?thread_ts=1717165521.276339&cid=C027XP5BTGB)

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -57,7 +57,7 @@ export const preprocess = async ({
   const start = Date.now()
   /** @type import('./typings.js').RawMeasurement[] */
   const measurements = await pRetry(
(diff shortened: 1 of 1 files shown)


bajtos (commented):
👍🏻 

@juliangruber Have you observed the impact of this change? Did it reduce the frequency of errors where we would receive an empty block response?

juliangruber:
Papertrail doesn't allow us to query velocity graphs for the last 4 days, and we're not capturing retries in Sentry, so unfortunately no. There are no related errors in Sentry.

PR #273, opened 2024-06-26
feat: show progress in discovering miner measurements

bajtos:
The script `bin/fetch-recent-miner-measurements.js` takes very long to complete. Usually, I abort it after collecting enough measurements.

This change adds a log message with the number of miner measurements found in each round. This makes it easier to know when we have collected enough measurements and can abort the script.


juliangruber (commented):
+1

PR #278, opened 2024-07-01
feat: add materialized view for top stations by measurement count

PatrickNercessian:
Link: https://github.com/space-meridian/roadmap/issues/118

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ import * as Sentry from '@sentry/node'
 import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { RoundData } from './lib/round.js'
+import { updateTopMeasurementStations } from './lib/platform-stats.js'
 import timers from 'node:timers/promises'
 
 // Tweak this value to improve the chances of the data being available
@@ -137,6 +138,10 @@ export const startEvaluate = async ({
       Sentry.captureException(err)
     })
   })
+
+  // Update top measurement stations every 12 hours
+  const pgClient = await createPgClient()
+  setInterval(async () => { await updateTopMeasurementStations(pgClient) }, 1000 * 60 * 60 * 12)
 
   while (true) {
     await timers.setTimeout(10_000)
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@ import * as Sentry from '@sentry/node'
 import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { RoundData } from './lib/round.js'
+import { updateTopMeasurementStations } from './lib/platform-stats.js'
 import timers from 'node:timers/promises'
 
 // Tweak this value to improve the chances of the data being available
@@ -139,10 +140,7 @@ export const startEvaluate = async ({
   })
 
   // Update top measurement stations every 12 hours
-  const pgClient = await createPgClient()
-  setInterval(async () => {
-    await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
-  }, 1000 * 60 * 60 * 12)
+  setInterval(async () => { await updateTopMeasurementStations(createPgClient) }, 1000 * 60 * 60 * 12)
 
   while (true) {
     await timers.setTimeout(10_000)
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@ import * as Sentry from '@sentry/node'
 import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { RoundData } from './lib/round.js'
-import { updateTopMeasurementStations } from './lib/platform-stats.js'
+import { updateTopMeasurementParticipants } from './lib/platform-stats.js'
 import timers from 'node:timers/promises'
 
 // Tweak this value to improve the chances of the data being available
@@ -140,7 +140,7 @@ export const startEvaluate = async ({
   })
 
   // Update top measurement stations every 12 hours
-  setInterval(async () => { await updateTopMeasurementStations(createPgClient) }, 1000 * 60 * 60 * 12)
+  setInterval(() => updateTopMeasurementParticipants(createPgClient), 1000 * 60 * 60 * 12)
 
   while (true) {
     await timers.setTimeout(10_000)
--- a/index.js
+++ b/index.js
@@ -3,7 +3,6 @@ import * as Sentry from '@sentry/node'
 import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { RoundData } from './lib/round.js'
-import { updateTopMeasurementStations } from './lib/platform-stats.js'
 import timers from 'node:timers/promises'
 
 // Tweak this value to improve the chances of the data being available
@@ -141,7 +140,9 @@ export const startEvaluate = async ({
 
   // Update top measurement stations every 12 hours
   const pgClient = await createPgClient()
-  setInterval(async () => { await updateTopMeasurementStations(pgClient) }, 1000 * 60 * 60 * 12)
+  setInterval(async () => {
+    await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
+  }, 1000 * 60 * 60 * 12)
 
   while (true) {
     await timers.setTimeout(10_000)
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -1,6 +1,5 @@
 import assert from 'node:assert'
 import createDebug from 'debug'
-import pg from 'pg'
 
 const debug = createDebug('spark:platform-stats')
 
@@ -166,11 +165,4 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
 
   assert.strictEqual(created.length, newAddresses.length)
   return ids.concat(created.map(r => r.id))
-}
-
-/**
- * @param {pg.Client} pgClient
- */
-export const updateTopMeasurementStations = async (pgClient) => {
-  await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
 }
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -1,5 +1,7 @@
 import assert from 'node:assert'
 import createDebug from 'debug'
+import * as Sentry from '@sentry/node'
+import pRetry from 'p-retry'
 
 const debug = createDebug('spark:platform-stats')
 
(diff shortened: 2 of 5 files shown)


migrations/011.do.top-measurements-stations-mv.sql
bajtos, line 1: Could you please use `012` as the index? I am adding `011` in https://github.com/filecoin-station/spark-evaluate/pull/277 and would like to land it first.

migrations/011.do.top-measurements-stations-mv.sql
```diff
@@ -0,0 +1,10 @@
+CREATE MATERIALIZED VIEW top_measurement_stations_mv AS
```
bajtos, line 1: Can we please find a name that conveys that these stats are for one day only?

The rows are keyed by `participant_address`, so perhaps we should use "participants" and not "stations" in the name.

migrations/011.do.top-measurements-stations-mv.sql
```diff
@@ -0,0 +1,10 @@
...
+SELECT
+  participant_address,
+  COUNT(DISTINCT inet_group) AS inet_group_count,
+  COUNT(DISTINCT station_id) AS station_count,
+  SUM(accepted_measurement_count) AS accepted_measurement_count
+FROM daily_stations
+WHERE day = CURRENT_DATE - INTERVAL '1 day'
+GROUP BY participant_address
+ORDER BY accepted_measurement_count DESC;
```
bajtos, line 10: Missing EOL at EOF.

```suggestion
ORDER BY accepted_measurement_count DESC;

```

bajtos (changes requested):
Nice start!

IIRC, spark-stats has read-only access to the DB; it's connected to a reader replica to reduce the load on the main DB server instance.

We need to move the query that updates the materialised view to spark-evaluate (i.e. to this pull request).

The data in the materialised view changes at midnight UTC, so it's best to run the update query soon after that.

A cron job would be ideal. Setting up cron jobs on Fly.io is a bit involved (see [their docs](https://fly.io/docs/blueprints/supercronic/)), so it may be easier to trigger the update using a timer, as you already did in the spark-stats pull request.

PatrickNercessian:
Addressed changes above @bajtos 

Do you think we need to unit test the refresh here too, or is it fine in spark-stats? Will require some refactoring here since we can't easily insert "yesterday" data into daily_stations as-is.

PatrickNercessian:
Also, what do you think about the location of the refresh? It could be in bin/, or index.js as-is now, or some other non-timer trigger, like on each round-starts.

index.js
```diff
@@ -138,6 +139,10 @@ export const startEvaluate = async ({
     })
   })
 
+  // Update top measurement stations every 12 hours
+  const pgClient = await createPgClient()
+  setInterval(async () => { await updateTopMeasurementStations(pgClient) }, 1000 * 60 * 60 * 12)
```
bajtos: I am concerned about keeping a long-living client connection. What happens when the connection is reset?

I propose to open & close a new client for every update. I think we should also add `pRetry` to handle intermittent errors.

```js
// here
setInterval(() => updateTopMeasurementStations(createPgClient), 1000 * 60 * 60 * 12)

// implementation - please refactor/clean up my spaghetti mess
export const updateTopMeasurementStations = async (pgClient) => {
  try {
    await pRetry(async () => {
      const pgClient = await createPgClient()
      try {
        await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_stations_mv')
      } finally {
        pgClient.close() // or end() ?
      }
   })
  } catch (err) {
    console.log(err)
    Sentry.captureException(err)
  }
}
```

lib/platform-stats.js
```diff
@@ -1,5 +1,6 @@
 import assert from 'node:assert'
 import createDebug from 'debug'
+import pg from 'pg'
```
bajtos: This is not used anywhere at runtime, is it? You should import `pg` inside the type declaration only.

migrations/012.do.top-measurements-stations-mv.sql
```diff
@@ -1,4 +1,4 @@
-CREATE MATERIALIZED VIEW top_measurement_stations_mv AS
+CREATE MATERIALIZED VIEW top_measurement_particpants_yesterday_mv AS
```
bajtos, line 1: Typo:

```suggestion
CREATE MATERIALIZED VIEW top_measurement_participants_yesterday_mv AS
```

index.js
bajtos: I'd like to discuss how much up-to-date we want to have this materialised view.

In the current proposal, we refresh the view every 12 hours. If the service process is started by Fly.io at 11.50 am UTC, we will refresh the view every day at 11.50 am UTC and 11.50 pm UTC. For most of the morning in Europe, the view will be showing data from the day before yesterday.

Is that good enough for this initial version? @juliangruber 

The alternative I am thinking of:
- At the process startup, if the time is between 0:00-0:30 am UTC, then run the update immediately.
- Always schedule the next update to be executed on the next day at 0:30 am UTC.

Thoughts?

index.js
```diff
@@ -138,6 +138,12 @@ export const startEvaluate = async ({
     })
   })
 
+  // Update top measurement stations every 12 hours
+  const pgClient = await createPgClient()
+  setInterval(async () => {
+    await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
+  }, 1000 * 60 * 60 * 12)
```
bajtos, line 145: > what do you think about the location of the refresh? It could be in bin/, or index.js as-is now, or some other non-timer trigger, like on each round-starts.

I think it's okay to have this code inside `startEvaluate()` function. It already installs other event listeners.

Just make sure `npm test` exits cleanly.

migrations/012.do.top-measurements-stations-mv.sql
PatrickNercessian: Wow, good catch!

lib/platform-stats.js
PatrickNercessian: Gotcha. I have since moved this to just index.js by the way, I think this is the older commit

bajtos:
> Do you think we need to unit test the refresh here too, or is it fine in spark-stats? Will require some refactoring here since we can't easily insert "yesterday" data into daily_stations as-is.

Let's add the tests to this repository, please. If the implementation and the tests live in different repositories, then it's difficult to update them together.

As for inserting "yesterday" data - that should be easy to achieve by calling `INSERT INTO daily_stations` directly from the tests, right?

See this test helper we already have for inserting data to `daily_participants`, I think you can use the same approach for your work.

https://github.com/filecoin-station/spark-evaluate/blob/4071d15cf01ed56eedaf050b840827f341467c16/test/helpers/queries.js#L10-L20

index.js
juliangruber: I believe this is good enough, let's iterate 👍

index.js
```diff
@@ -139,10 +140,7 @@ export const startEvaluate = async ({
   })
 
   // Update top measurement stations every 12 hours
-  const pgClient = await createPgClient()
-  setInterval(async () => {
-    await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
-  }, 1000 * 60 * 60 * 12)
+  setInterval(async () => { await updateTopMeasurementStations(createPgClient) }, 1000 * 60 * 60 * 12)
```
bajtos, line 143: Nitpick: there is no need to wrap the function call in another async function and `await`.

```suggestion
  setInterval(() => updateTopMeasurementStations(createPgClient), 1000 * 60 * 60 * 12)
```

lib/platform-stats.js
```diff
@@ -166,3 +170,27 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
   assert.strictEqual(created.length, newAddresses.length)
   return ids.concat(created.map(r => r.id))
 }
+
+/**
+ * @param {import('./typings.js').CreatePgClient} createPgClient
+ */
+export const updateTopMeasurementStations = async (createPgClient) => {
```
bajtos, line 177:
```suggestion
export const updateTopMeasurementParticipants = async (createPgClient) => {
```

... to match the name we use in other places.

index.js
bajtos: ok 👍🏻

lib/platform-stats.js
```diff
@@ -11,15 +13,16 @@ const debug = createDebug('spark:platform-stats')
...
 }
 
 /**
  * @param {import('pg').Client} pgClient
+ * @param {Date} day
  * @param {import('./preprocess.js').Measurement[]} honestMeasurements
  * @param {import('./preprocess.js').Measurement[]} allMeasurements
  */
-export const updateDailyStationStats = async (pgClient, honestMeasurements, allMeasurements) => {
+export const updateDailyStationStats = async (pgClient, day, honestMeasurements, allMeasurements) => {
```
bajtos, line 25: I have mixed feelings about introducing the `today` parameter to `updateDailyStationStats()`, but I also don't have any good counter-arguments, so let's give this a try and see how it works in practice.

bajtos (approved):
LGTM with a few minor comments to consider.

@juliangruber, could you please review the changes too?

index.js
juliangruber: Agreed

lib/platform-stats.js
juliangruber: +1

lib/platform-stats.js
juliangruber: Since it's always today in production, and only a different day in tests, can we add this as an optional parameter instead?

juliangruber (changes requested):

lib/platform-stats.js
PatrickNercessian: Ah missed that, addressed in https://github.com/filecoin-station/spark-evaluate/pull/278/commits/3b8b7416a86cb7cfe21fd52ca6f45c8c8b8c6f08

juliangruber (approved):

lib/platform-stats.js
PatrickNercessian: I also didn't love it at first, but the alternative is fully mimicking the entire updateDailyStationStats function with control over the date to be used in tests (because it has somewhat nuanced logic compared to other insertion-handling functions we have)

Used an optional parameter in https://github.com/filecoin-station/spark-evaluate/pull/278/commits/3b8b7416a86cb7cfe21fd52ca6f45c8c8b8c6f08

index.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/278/commits/3b8b7416a86cb7cfe21fd52ca6f45c8c8b8c6f08

bajtos:
🎉

PR #295, opened 2024-07-30
test: update integration test to use round 12012

bajtos:
- **test: add cache for fetched measurements**
- **test: update integration test to use round 12012**

Links:
- https://github.com/filecoin-station/spark-evaluate/pull/287
- https://github.com/filecoin-station/spark/issues/27


test/integration.test.js
```diff
@@ -417,3 +429,19 @@ const debugDumpData = (name, data) => {
   for (const k of keys) sorted[k] = data[k]
   debug(name, JSON.stringify(sorted, null, 2))
 }
+
+const fetchMeasurementsWithCache = async (cid) => {
```
bajtos, line 433: I copied this function from [bin/dry-run.js](https://github.com/filecoin-station/spark-evaluate/blob/53c7923ded7d398266b97491c583e3da5d2621f8/bin/dry-run.js).

If you think it's worth extracting this function into a shared helper, then let me know, and I'll open a follow-up pull request.

juliangruber: Fine with me as is 👍 It's small in scope and not complex

juliangruber (approved):
Code changes lgtm, but please remind me why we need to update the round to use?

bajtos:
> Code changes lgtm, but please remind me why we need to update the round to use?

For the deterministic task assignment, we need to map Spark round to DRAND's epoch. We do so by mapping the Filecoin epoch (block number) when the Spark started to a point in time, and then mapping that point in time to DRAND epoch.

We started tracking Spark round start epoch relatively recently, we don't have it for the round 3602n. spark-api returns `startEpoch: 0` and drand-client fails with the following error:

```
Error: Cannot request a round before the genesis time
```

test/integration.test.js
```diff
@@ -111,8 +116,8 @@ describe('preprocess-evaluate integration', () => {
     assert.match(fetchMeasurementsDuration, /^\d+i/)
     assert.deepStrictEqual(preprocessingStats, {
       round_index: `${MERIDIAN_ROUND}i`,
-      total_measurements: '15889i',
-      valid_measurements: '15889i'
+      total_measurements: '44677i',
+      valid_measurements: '44677i'
```
bajtos, line 120: This is interesting. Spark is recording ~3x more measurement now.

PR #298, opened 2024-07-31
fix: calculate committee stats from honest measurements

bajtos:
When calculating number of participants/subnets/measurements per committee (per retrieval task), we must use only accepted measurements.

When using all measurements, we include data from measurements for tasks (committees) that are not valid for this round, which distorts the statistics. For example, the minimum number of participants per committee is much smaller.


juliangruber (approved):
makes sense!

PR #304, opened 2024-08-01
feat: form committees and find majority results

bajtos:
Group accepted measurements to committees on a per-retrieval-task basis. Evaluate each committee to find an absolute majority result. Reject results that are in minority.
    
When the committee is too small to give us confidence in a majority being honest, or if we cannot find an absolute majority, then reject all measurements in such committee.

Links:
- https://github.com/filecoin-station/spark-evaluate/pull/303
- https://github.com/filecoin-station/spark-evaluate/pull/302
- https://github.com/filecoin-station/spark-evaluate/pull/306

TODO:
- cross-check more fields
  - [x] CAR size in bytes
  - [x] CAR checksum
- [x] add telemetry to monitor committees too small or committees with no absolute majority

OUT OF SCOPE
- Change the way how we calculate retrieval success rates from _per measurement_ to _per deal_.

NOTE
We are not cross-checking the fields that were used to calculate the retrievalResult value:
  - Response status code
  - Was it a timeout?
  - Was the CAR too large?

If there is an agreement on the retrieval result, then those fields must have the same value, too.


lib/committee.js
```diff
@@ -0,0 +1,143 @@
...
+
+    if (debug.enabled) {
+      debug('- %s values found: %o', fieldName, Object.fromEntries(countsOfValues.entries()))
+    }
+
+    const majorityValue = uniqueValues.pop()
+    const majoritySize = countsOfValues.get(majorityValue)
+    debug('- %s majority=%s committee-size=%s value=%o', fieldName, majoritySize, committeeSize, majorityValue)
+    if (majoritySize <= committeeSize / 2) {
+      this.#checkFailed('MAJORITY_NOT_FOUND')
```
bajtos, line 101: Are we ok to use one code `MAJORITY_NOT_FOUND` for all fields, or we want more granular codes that will allow us to understand for which field the majority was not found?

E.g. `MAJORITY_NOT_FOUND_FOR_indexerResult`, `MAJORITY_NOT_FOUND_FOR_providerResult`, etc.

lib/committee.js
```diff
@@ -0,0 +1,143 @@
...
+   * @template {keyof Measurement} FIELD_NAME
+   * @param {FIELD_NAME} fieldName
+   */
+  #checkMeasuredField (fieldName) {
+    /** @type {Map<Measurement[FIELD_NAME], number>} */
+    const countsOfValues = new Map()
+    let committeeSize = 0
+    for (const m of this.measurements) {
+      if (m.fraudAssessment !== 'OK') continue
+      committeeSize++
```
bajtos, line 86: I think I got this wrong.

Let's consider the following case:

- The committee has 30 measurements
- 20 measurements agree on the same `providerId`
- From those 20 measurements above, 12 agree on the `retrievalResult`
- In the current proposal, we calculate `committeeSize=20` and `majoritySize=12` and conclude that we have a majority we can trust.

That has two problems:
1. We need at least 30 measurements to give us "honest majority". With a committee of 20 measurements, we can no longer be sure about honesty.
2. We will apply several checks; each check can halve the committee size. Towards the end, we can end up with committees containing only a handful of measurements, and we would still treat the majority result as meaningful. That's clearly wrong.

I'll will rework this code to always use the original committee size when determining if we have an absolute majority; and I will also add a test for this.

@patrickwoodhead You know a bit about statistics too, do you agree with my reasoning?

lib/committee.js
juliangruber: Good question! Is it an actionable error? And does thie mean the whole measurement is rejected if one field is out of majority?

juliangruber: I thought the idea was to use committees in order to distinguish "retrieval works" from "retrieval doesn't work", from that perspective we don't need more granularity

lib/committee.js
juliangruber: This is the first time I'm thinking about individual fields wrt to committees, I thought they were only concerned with "rsr pass" vs "rsr fail". Does it help the design to form committees on every field? (Not saying it doesn't, just asking).

If we're forming a committee on each field, I think this class needs to be refactored so there's one instance per field, as there really is one committee per field.

lib/evaluate.js
```diff
@@ -366,6 +369,27 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
     }
   }
 
+  //
+  // 3. Group measurements to per-task committees and find majority result
```
juliangruber, line 373: This might change to `per-task per-field committees`, depending on how discussion above goes

juliangruber (changes requested):

lib/committee.js
bajtos: It's not actionable, but it gives us a better understanding of where the measurements disagree.

> I thought the idea was to use committees in order to distinguish "retrieval works" from "retrieval doesn't work", from that perspective we don't need more granularity

👍🏻 

We can easily add more granularity later if we need it.

lib/committee.js
bajtos: > I think this is indeed problematic, as the result depends on the order the checks have been performed. Each check should always test all measurements instead of checks influencing each other

I think that's not going to work.

The retrieval check has the following steps:
1. Resolve miner ID like `f0...` to provider ID (a libp2p peer id)
2. Query IPNI to find provider records for the CID tested and pick only the records advertised by the provider ID found in the previous step. If no records are found, then abort the algorithm.
3. Use the information in the provider record (multiaddr, protocol) to retrieve the content
4. Calculate checksum of the received CAR file

If nodes resolve `minerId` to different `ProviderID` values, then in the next step - IPNI lookup - they will pick different advertisements and perform the retrieval using different protocols & different hosts to retrieve from.

In your proposal, we will look at the `retrievalResult` field in isolation, ignoring how the retrieval was performed.

The way how I look at majorities: we want to find a majority that reports the same tuple `(providerId, indexerResult, retrievalResult)` - that's what I implemented in this PR. In the future, I want to extend that tuple to include also other fields like `address`, `protocol`, `car_checksum`, etc.

lib/committee.js
```diff
@@ -0,0 +1,143 @@
...
+    // - protocol,
+
+    if (!this.#checkMeasuredField('retrievalResult')) return
+
+    // TODO
+    // - status_code,
+    // - timeout,
+    // - car size
+    // - car checksum
+    // - car too large
```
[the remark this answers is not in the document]
bajtos: I agree the side effects are not great. However, your solution does not completely remove the need for side effects. If we find the majority, then we still need to modify the minority measurements.

Also, your proposal will make it difficult to interactively debug this function because you cannot step over individual checks in the condition. Unless the debugger can do that now?

lib/committee.js
juliangruber: What about forming a compound key and using that to find a majority? This way it's a single step process. Eg:

```js
  const key = `${m.providerId}${m.indexerResult}${m.retrievalResult}...`
  counts[key]++
}
```

lib/committee.js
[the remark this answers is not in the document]
juliangruber: > I agree the side effects are not great. However, your solution does not completely remove the need for side effects. If we find the majority, then we still need to modify the minority measurements.

With my proposal above this multi step process might not be necessary any more

> Also, your proposal will make it difficult to interactively debug this function because you cannot step over individual checks in the condition. Unless the debugger can do that now?

The debugger should be able to do that yeah

lib/committee.js
bajtos: We could do that. We will need multiple compound steps, though. One step per the output field we want to derive from the measurements:

- indexerResult requires `(providerId, indexerResult)`
- retrievalResult requires `(providerId, indexerResult, retrievalResult,...)`


There is some duplication between these steps. It's not a problem if we have two steps, but it will become a problem if we add more output fields.

My main concern about this compound-key-based approach is the reduced visibility into the underlying data. 

For example, in my proposal, after we add a check for protocol & provider address, we will be able to tell that the majority agrees on the provider ID and the indexer result, but the committee does not agree on the provider address field.

If we use a compound key, we will know that the majority agrees on the provider ID and the indexer result - thanks to the first step. We won't know why the committee does not agree on the overall outcome.

Another example: car_checksums. In my proposal, we will see that a majority agrees on the provider address and protocol and that the content can be retrieved, but the measurements report different CAR checksums. In your proposal, we would need to analyse the measurements manually to find that information.

Having said that: In https://github.com/filecoin-station/spark-evaluate/pull/304/files#diff-ee367bc5e91f746457be9f758a7f69be7f0bf930e08904cd535cbd3b5e93d5e8, we agreed to use a single coarse-grained code `MAJORITY_NOT_FOUND`. That means we won't see the details I described above in the telemetry and stats. We would get them only when running a dry run with DEBUG logs enabled.

In that light, maybe this reduced visibility is a somewhat lesser problem.

lib/committee.js
```diff
@@ -0,0 +1,143 @@
...
+    this.retrievalResult = 'MAJORITY_NOT_FOUND'
+  }
+
+  /**
+   * @param {number} requiredCommitteeSize
+   * @returns
+   */
+  evaluate (requiredCommitteeSize = REQUIRED_COMMITTEE_SIZE) {
+    debug(
+      'Evaluating task %o with %s accepted measurements',
```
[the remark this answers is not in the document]
bajtos: How about this?

```suggestion
      'Evaluating task %o with a committee of %s measurements',
```

lib/committee.js
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * @template {keyof Measurement} FIELD_NAME
+   * @param {FIELD_NAME} fieldName
+   */
+  #checkMeasuredField (fieldName) {
+    /** @type {Map<Measurement[FIELD_NAME], number>} */
+    const countsOfValues = new Map()
+    let committeeSize = 0
+    for (const m of this.measurements) {
+      if (m.fraudAssessment !== 'OK') continue
```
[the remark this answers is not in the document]
bajtos: This is tricky, which is not ideal.

This class expects to receive only measurements that passed previous evaluation steps, i.e. the honest/accepted measurements only.

However, each `check` step can change the evaluation of some measurements from "OK" to "MINORITY_RESULT". The following steps need to ignore those rejected measurements per my arguments in https://github.com/filecoin-station/spark-evaluate/pull/304#discussion_r1700280538

lib/committee.js
juliangruber: > We could do that. We will need multiple compound steps, though. One step per the output field we want to derive from the measurements:
> 
> * indexerResult requires `(providerId, indexerResult)`
> * retrievalResult requires `(providerId, indexerResult, retrievalResult,...)`
> 
> There is some duplication between these steps. It's not a problem if we have two steps, but it will become a problem if we add more output fields.

I think as long as we only need `MAJORITY_NOT_FOUND` this is no problem as one step of checking `(rpoviderId, indexerResult, retrievalResult)` together will cover all cases.

I'm concerned that this multi step process isn't only more complex but also provides results that are harder to trust: Say measurements disagree on two results (indexerResult and retrievalResult), the order in which we execute the checks will tell which result we get. In this case this isn't a problem, because if the indexer result is different, the retrieval result doesn't matter. But are we sure we will always add results that are strictly sequential?

juliangruber: Are these stepped per-field committees part of the new Spark spec?

bajtos: > I'm concerned that this multi step process isn't only more complex but also provides results that are harder to trust: Say measurements disagree on two results (indexerResult and retrievalResult), the order in which we execute the checks will tell which result we get. In this case this isn't a problem, because if the indexer result is different, the retrieval result doesn't matter. But are we sure we will always add results that are strictly sequential?

The idea behind this current multi-step is to mirror the steps that Spark checker nodes perform. We verify that after each step of the retrieval testing protocol, we have a majority that agrees on what should have been the outcome of that step.

But I agree with you that this design makes it more difficult to reason about the correctness.

I will try the compound-key alternative and see if I run into any issues.

> Are these stepped per-field committees part of the new Spark spec?

No, I don't think we have any spec for how exactly committees should be formed.

lib/committee.js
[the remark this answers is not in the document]
juliangruber: lgtm!

lib/committee.js
juliangruber: Then I believe we need to do that first, before jumping into code, or keep the committee logic as simple as possible

bajtos:
I compared the output of `node bin/dry-run.js 12649`.

- Only 1.3% of measurements have a minority result
- Only 0.1% of committees are too small (less than 30 measurements)

```diff
-EVALUATE ROUND 12649n: built per-node task lists in 3804ms [Tasks=1000;TN=15;Nodes=55235]
-EVALUATE ROUND 12649n: added 2980n as rounding to MAX_SCORE
-EVALUATE ROUND 12649n: Evaluated 333523 measurements, found 93814 honest entries.
+EVALUATE ROUND 12649n: built per-node task lists in 3958ms [Tasks=1000;TN=15;Nodes=55235]
+EVALUATE ROUND 12649n: added 3871n as rounding to MAX_SCORE
+EVALUATE ROUND 12649n: Evaluated 333523 measurements, found 92574 honest entries.
 {
-  OK: 93814,
+  OK: 92574,
   TASK_NOT_IN_ROUND: 37557,
   DUP_INET_GROUP: 34102,
   TOO_MANY_TASKS: 81874,
-  TASK_WRONG_NODE: 86176
+  TASK_WRONG_NODE: 86176,
+  MINORITY_RESULT: 1213,
+  COMMITTEE_TOO_SMALL: 27
 }
```

The overall RSR went slightly up from 5.29% to 5.31%.

```diff
- success_rate: '0.052934530027501224',
+ success_rate: '0.053114265344481174',
```

The number of IPNI 404 errors decreased slightly from 61.08% to 60.09%.

```diff
-  result_rate_IPNI_ERROR_404: '0.6107830387788603',
+  result_rate_IPNI_ERROR_404: '0.6089398751269255',
```

The CPU & memory used by the evaluation did not change much.

```diff
-Duration: 10114ms
+Duration: 10630ms
 {
-  rss: 450183168,
+  rss: 451018752,
```

bajtos:
@juliangruber the PR is ready for another round of review.

> cross-check more fields

This should be trivial to add - update the array of fields checked + write more tests.

lib/committee.js
juliangruber: I think we need to nail down the conceptual side first before there's more merit in reviewing the code. Otherwise we can decide to change the concept and that causes code rewrites and then that review was mostly meaningless.

lib/committee.js
```diff
@@ -0,0 +1,200 @@
...
+    )
+
+    if (this.#measurements.length < requiredCommitteeSize) {
+      debug('→ committee is too small (size=%s required=%s); retrievalResult=COMMITTEE_TOO_SMALL',
+        this.#measurements.length,
+        requiredCommitteeSize
+      )
+      this.indexerResult = 'COMMITTEE_TOO_SMALL'
+      this.retrievalResult = 'COMMITTEE_TOO_SMALL'
+      this.#measurements.forEach(m => { m.fraudAssessment = 'COMMITTEE_TOO_SMALL' })
```
juliangruber, line 70: nit:

```suggestion
      for (const m of this.#measurements) m.fraudAssessment = 'COMMITTEE_TOO_SMALL'
```

lib/committee.js
```diff
@@ -0,0 +1,200 @@
...
+      // - car checksum
+      // - car too large
+    ]
+
+    const retrievalResultMajority = this.#findMajority(retrievalResultProps)
+    if (retrievalResultMajority) {
+      this.retrievalResult = retrievalResultMajority.majorityValue.retrievalResult
+      retrievalResultMajority.minorityMeasurements.forEach(m => {
+        m.fraudAssessment = 'MINORITY_RESULT'
+      })
```
juliangruber, line 105: nit:

```suggestion
      for (const m of retrievalResultMajority.minorityMeasurements) {
        m.fraudAssessment = 'MINORITY_RESULT'
      }
```

lib/committee.js
```diff
@@ -0,0 +1,200 @@
...
+    } else {
+      this.retrievalResult = 'MAJORITY_NOT_FOUND'
+      this.#measurements.forEach(m => { m.fraudAssessment = 'MAJORITY_NOT_FOUND' })
```
juliangruber, line 108: nit:

```suggestion
      for (const m of this.#measurements) m.fraudAssessment = 'MAJORITY_NOT_FOUND'
```

lib/evaluate.js
```diff
@@ -368,6 +394,22 @@ export const runFraudDetection = async (roundIndex, measurements, sparkRoundDeta
...
+  //
+  const iterateHonestMeasurements = function * () {
+    for (const m of measurements) {
+      if (m.fraudAssessment !== 'OK') continue
+      yield m
+    }
+  }
+
+  const committees = groupMeasurementsToCommittees(iterateHonestMeasurements())
```
juliangruber, line 407: Could this be simpler like this?

```suggestion
  const committees = groupMeasurementsToCommittees(
    measurements.filter(m => m.fraudAssessment === 'OK')
  )
```

bajtos: I didn't want to duplicate the array of measurements because there are hundreds of thousands of them. Do you think it's a preliminary optimisation to avoid now?

juliangruber: If you didn't measure it then yes I think so

juliangruber: I did a small test:

```js
{ const a = []; for (let i = 0; i < 1_000_000; i++) a.push({ beep: 'boop' }); const s = new Date(); [...a]; console.log(new Date() - s) }
52
```

52 milliseconds isn't little, especially with a way smaller object size. I'm +1 for this optimization, if we document it :)

lib/committee.js
bajtos: Marking this thread as resolved, the PR uses the compound-key approach now.

lib/evaluate.js
bajtos: SGTM, added a code comment in d61cef9

lib/committee.js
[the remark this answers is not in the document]
bajtos: Marking this thread as resolved, the PR uses the compound-key approach now.

lib/committee.js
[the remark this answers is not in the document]
bajtos: Marking this thread as resolved, the PR uses the compound-key approach now and the method `addMeasurement` asserts that each measurement has `fraudAssessment` value `OK`.

lib/committee.js
```diff
@@ -0,0 +1,200 @@
...
+  constructor ({ cid, minerId }) {
+    this.retrievalTask = { minerId, cid }
+
+    this.#measurements = []
+
+    /** @type {string  | CommitteeCheckError} */
+    this.indexerResult = 'MAJORITY_NOT_FOUND'
+
+    /** @type {RetrievalResult} */
+    this.retrievalResult = 'MAJORITY_NOT_FOUND'
```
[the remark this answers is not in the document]
juliangruber: +1 to having the result be `undefined` or something else, which we can differentiate from the previous default state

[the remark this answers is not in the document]
juliangruber: And I'm not sure whether removing the class does help, let's see how it goes?

bajtos:
Comparing the output of `node bin/dry-run.js 13944` - `main` vs `b8916a9`:

```diff
-EVALUATE ROUND 13944n: built per-node task lists in 845ms [Tasks=1000;TN=15;Nodes=12334]
-EVALUATE ROUND 13944n: added 3920n as rounding to MAX_SCORE
-EVALUATE ROUND 13944n: Evaluated 215182 measurements, found 98111 honest entries.
+EVALUATE ROUND 13944n: built per-node task lists in 865ms [Tasks=1000;TN=15;Nodes=12334]
+EVALUATE ROUND 13944n: added 3931n as rounding to MAX_SCORE
+EVALUATE ROUND 13944n: Evaluated 215182 measurements, found 97335 honest entries.
 {
-  OK: 98111,
+  OK: 97335,
   TASK_NOT_IN_ROUND: 16191,
   DUP_INET_GROUP: 24871,
   TOO_MANY_TASKS: 75847,
-  TASK_WRONG_NODE: 162
+  TASK_WRONG_NODE: 162,
+  MINORITY_RESULT: 573,
+  MAJORITY_NOT_FOUND: 203
 }
```

bajtos:
Comparing the output of `node bin/dry-run.js 13944` - `main` vs `b8916a9`:
- Only 0.26% (573/215182) of measurements have a minority result
- All committees were large enough to give us the confidence of the majority being honest

```diff
-EVALUATE ROUND 13944n: built per-node task lists in 845ms [Tasks=1000;TN=15;Nodes=12334]
-EVALUATE ROUND 13944n: added 3920n as rounding to MAX_SCORE
-EVALUATE ROUND 13944n: Evaluated 215182 measurements, found 98111 honest entries.
+EVALUATE ROUND 13944n: built per-node task lists in 865ms [Tasks=1000;TN=15;Nodes=12334]
+EVALUATE ROUND 13944n: added 3931n as rounding to MAX_SCORE
+EVALUATE ROUND 13944n: Evaluated 215182 measurements, found 97335 honest entries.
 {
-  OK: 98111,
+  OK: 97335,
   TASK_NOT_IN_ROUND: 16191,
   DUP_INET_GROUP: 24871,
   TOO_MANY_TASKS: 75847,
-  TASK_WRONG_NODE: 162
+  TASK_WRONG_NODE: 162,
+  MINORITY_RESULT: 573,
+  MAJORITY_NOT_FOUND: 203
 }
```

The overall RSR went slightly _down_ from 8.72% to 8.65%.

```diff
-  success_rate: '0.0871869617066384',
+  success_rate: '0.08654646324549237',
```

The number of IPNI 404 errors _increased_ slightly from `59.87%` to `60.35%`.

```diff
- result_rate_IPNI_ERROR_404: '0.5986892397386634',
+ result_rate_IPNI_ERROR_404: '0.6034622694816869',
```

The CPU & memory used by the evaluation did not change much.

```diff
-Duration: 5148ms
+Duration: 5551ms
 {
-  rss: 370573312,
+  rss: 312492032,
```

juliangruber (approved):
Great work!

PR #306, opened 2024-08-02
test: remove the integration test

bajtos:
This test is evaluating one batch of real measurements. With the upcoming change introducing committees and majorities, a single batch of measurements is not enough to produce committees with the required size. As a result, most measurements are rejected with `COMMITTEE_TOO_SMALL` evaluation. In a recent test run, the number of accepted measurements dropped from 14,512 to 2,530.

We could fix this test by evaluating all batches in one round, but that would extremely slow down this test to the order of minutes.

Let's remove this test for now. If we find a need to test evaluation results of real data, then I'd like to rework the `dry-run` CI step to diff the dry-run output against a snapshot.

Links:
- https://github.com/filecoin-station/spark-evaluate/pull/304


juliangruber (approved):
reasoning sgtm!

PR #311, opened 2024-08-08
add new schema tables and accompanying write logic

PatrickNercessian:
Links:
* https://github.com/space-meridian/roadmap/issues/132


lib/platform-stats.js
```diff
@@ -165,37 +271,168 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+        SELECT
+          day,
+          SUM(accepted_measurement_count) AS accepted_measurement_count,
+          SUM(total_measurement_count) AS total_measurement_count,
+          COUNT(DISTINCT station_id) AS distinct_active_station_count,
+          COUNT(DISTINCT participant_id) AS distinct_participant_address_count
+        FROM
+          recent_station_details
+        WHERE
+          day <= $1::DATE
```
bajtos, line 342: As an alternative to calculating "two days ago" in JavaScript, I think we should be able to use something like `now() - INTERVAL '2 days'` or `now()::DATE - 'INTERVAL '2 days'`.

We need to be careful about time zones - 1 am on Tuesday in Prague is 7 pm on Monday in New York. I expect that all our services are running in UTC time, so we should be fine.

bajtos (commented):
The implementation looks great at the high level 👍🏻 

Remember to run `standard --fix`; there are many lines ending with a semicolon.

lib/platform-stats.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/bc1515657bc9a0209a1890df58c19734ec31d6ff

I also did the same for the lastMonth calculation

I agree that timezones should theoretically not mess us up here; I think we do similar things today with dashboard code

index.js
```diff
@@ -141,7 +141,7 @@ export const startEvaluate = async ({
   })
 
   // Update top measurement stations every 12 hours
-  setInterval(() => updateTopMeasurementParticipants(createPgClient), 1000 * 60 * 60 * 12)
+  setInterval(() => periodicDatabaseRefresh(createPgClient), 1000 * 60 * 60 * 12)
```
PatrickNercessian, line 144: Do we want this still at 12 hours? I think I had done it before with the leaderboard, so that our 24h cache didn't clear right before it, and then we'd have the same stale data for ~48 hours.

I might be thinking about it wrong though. We could leave it as-is, or switch to 24h, or have two separate intervals, 12h for the leaderboard update, and 24h for the others. What do we think?

lib/platform-stats.js
```diff
@@ -199,3 +316,119 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+/**
+ * @param {import('pg').Client} pgClient
+ */
+export const aggregateAndCleanupRecentData = async (pgClient) => {
+  try {
+    await pgClient.query('BEGIN')
+
+    // The ON CONFLICT theoretically should not happen, because it's two days old data,
+    // which should not be updated anymore. But it's a good practice to have it.
+    await pgClient.query(`
```
PatrickNercessian, line 329: Thoughts on this?

test/platform-stats.test.js
```diff
@@ -91,7 +99,7 @@ describe('platform-stats', () => {
           total_measurement_count
         FROM
           daily_stations
-        ORDER BY station_id`
+        ORDER BY station_id, participant_address, inet_group`
```
PatrickNercessian, line 102: This test is separate from the code changes, but it was randomly failing due to ordering issues

lib/platform-stats.js
```diff
@@ -165,29 +271,40 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+
+export const periodicDatabaseRefresh = async (createPgClient) => {
+  const pgClient = await createPgClient()
+  try {
+    await updateTopMeasurementParticipants(pgClient)
+    await aggregateAndCleanupRecentData(pgClient)
+    await updateMonthlyActiveStationCount(pgClient)
+  } finally {
+    await pgClient.end()
+  }
```
juliangruber, line 298: This implementation has the problem that if one function fails, the ones following won't run.

If we want these functions to run in series, something like this could work:

```js
let err

try {
  await fnA()
} catch (_err) {
  err = _err
}

// ...

await pgClient.end()
if (err) {
  throw err
}
```

If parallel execution is ok, this simpler version works:

```js
try {
  await Promise.allSettled([
    fnA(),
    // ...
  ])
} finally {
  await pgClient.end()
}
```

index.js
juliangruber: This depends on the data. Do you think the new functions should run on a different schedule? If there's no obvious issue I would suggest to keep it simple and iterate later.

lib/platform-stats.js
```diff
@@ -99,30 +100,135 @@ export const updateDailyStationStats = async (
...
+    participantIds
+  ])
+}
+
+/**
+ * @param {import('pg').Client} pgClient
+ * @param {import('./preprocess.js').Measurement[]} honestMeasurements
+ * @param {import('./preprocess.js').Measurement[]} allMeasurements
+ * @param {Map<string, number>} participantMap
+ * @param {Date} day
```
juliangruber, line 121: this parameter should be marked as optional

lib/platform-stats.js
```diff
@@ -99,30 +100,135 @@ export const updateDailyStationStats = async (
...
+  /** @type {(m: import('./preprocess.js').Measurement, participantId: number) => string} */
+  const getStationParticipantKey = (m, participantId) => JSON.stringify({
+    stationId: m.stationId,
+    participantId
+  })
+  /** @type {(m: import('./preprocess.js').Measurement, participantId: number) => string} */
+  const getParticipantSubnetKey = (m, participantId) => JSON.stringify({
+    participantId,
+    subnet: m.inet_group
+  })
```
juliangruber, line 143: These functions don't depend on the parent function's scope and can be moved out

lib/platform-stats.js
juliangruber: Thoughts on the `ON CONFLICT`, or the whole query?

lib/platform-stats.js
```diff
@@ -199,3 +316,119 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+ * @param {import('pg').Client} pgClient
+ */
+export const aggregateAndCleanupRecentData = async (pgClient) => {
+  try {
+    await pgClient.query('BEGIN')
+
+    // The ON CONFLICT theoretically should not happen, because it's two days old data,
+    // which should not be updated anymore. But it's a good practice to have it.
+    await pgClient.query(`
+      WITH rsdn_summary AS (
```
juliangruber, line 330: What does `rsdn` stand for?

lib/platform-stats.js
```diff
@@ -199,3 +316,119 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+          COUNT(DISTINCT station_id) AS distinct_active_station_count,
+          COUNT(DISTINCT participant_id) AS distinct_participant_address_count
+        FROM
+          recent_station_details
+        WHERE
+          day <= CURRENT_DATE - INTERVAL '2 days'
+        GROUP BY
+          day
+      ),
+      rps_summary AS (
```
juliangruber, line 344: What does `rps` stand for`?

migrations/013.do.dashboard-schema-migration.sql
```diff
@@ -0,0 +1,38 @@
...
+);
+
+CREATE TABLE daily_measurements_summary (
+    day DATE NOT NULL PRIMARY KEY,
+    accepted_measurement_count INTEGER NOT NULL,
+    total_measurement_count INTEGER NOT NULL,
+    distinct_active_station_count INTEGER NOT NULL,
+    distinct_participant_address_count INTEGER NOT NULL,
+    distinct_inet_group_count INTEGER NOT NULL
+);
```
juliangruber, line 38:
```suggestion
);

```

lib/platform-stats.js
PatrickNercessian: Ah I think this was a remnant of my local tests for several different schemas, will fix

lib/platform-stats.js
PatrickNercessian: `recent_participant_subnets` table

index.js
PatrickNercessian: I think it's fine to be on the same schedule, but it does mean that we're running them more often than necessary. Ideally, they should gracefully do nothing though. Now that I think about it, I should add some unit tests to make sure.

lib/platform-stats.js
PatrickNercessian: Ah good catch. I was planning to use Promise.allSettled before committing but it slipped my mind

PatrickNercessian:
Ah new issue discovered that doesn’t actually fail the tests. Basically it stems from an edge case where all measurements sent by a participant have never been “honest”, because [`mapParticipantsToIds` takes in just participants with honest measurements](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R14)

Therefore, [when we later try to push to `recent_station_details`](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R157-R165), the key has no participant ID, so we fail on a not-null constraint violation

PatrickNercessian:
Aside from the above comments from @juliangruber, here are next steps:
1. add a unit test for `periodicDatabaseRefresh`
2. add/enhance tests to also test whether additional calls to periodicDatabaseRefresh's sub-methods gracefully do nothing
3. fix edge case mentioned above

lib/platform-stats.js
juliangruber: ok, what do you think about just writing that out?

juliangruber:
> Ah new issue discovered that doesn’t actually fail the tests. Basically it stems from an edge case where all measurements sent by a participant have never been “honest”, because [`mapParticipantsToIds` takes in just participants with honest measurements](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R14)
> 
> Therefore, [when we later try to push to `recent_station_details`](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R157-R165), the key has no participant ID, so we fail on a not-null constraint violation
> 
> One possible fix would be to start sending the full list of measurements to `mapParticipantsToIds`.

Good question! Your proposed fix seems straight forward to me, as long as there are no resource concerns let's go with that

PatrickNercessian:
> > Ah new issue discovered that doesn’t actually fail the tests. Basically it stems from an edge case where all measurements sent by a participant have never been “honest”, because [`mapParticipantsToIds` takes in just participants with honest measurements](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R14)
> > Therefore, [when we later try to push to `recent_station_details`](https://github.com/filecoin-station/spark-evaluate/pull/311/files#diff-8ac46082eb9901bb44d0a0c1b3c9f2d6b66de8faa05d04bec47eb0acc65e4ff0R157-R165), the key has no participant ID, so we fail on a not-null constraint violation
> > One possible fix would be to start sending the full list of measurements to `mapParticipantsToIds`.
> 
> Good question! Your proposed fix seems straight forward to me, as long as there are no resource concerns let's go with that

I think it depends on how many extra participants we expect to start adding to the `participants` table by doing this (all the ones that have exclusively no honest measurements). I assume this is super small, but does this introduce a new "attack vector" or something else of concern? My intuition is that this is no worse than how we currently allow (and experience) a huge amount of station_id churn, but I want to make sure we're on the same page here.

juliangruber:
Oh wait I thought this was just about in memory state. If it's persisting in the database, I think we either:
- estimate that adding the extra participants is no problem
- don't process participants without any valid measurements
What do you think about both?

PatrickNercessian:
> Oh wait I thought this was just about in memory state. If it's persisting in the database, I think we either:
> * estimate that adding the extra participants is no problem
> * don't process participants without any valid measurements
> 
> What do you think about both?

1. My thought is that this is no problem, at least no worse than the station_id churn problem.
2. As in, we just catch that specific error and do nothing? I think this would be fine code-wise, but then we will have an incomplete dataset of "total_measurement_count".

My light preference would (1), as long as you also can't think of a reason for it to become a problem.

juliangruber:
> My thought is that this is no problem, at least no worse than the station_id churn problem.

Adding another problem of the same severity as station_id sounds like a bad idea :P

> As in, we just catch that specific error and do nothing? I think this would be fine code-wise, but then we will have an incomplete dataset of "total_measurement_count".

Hm no I think it would need to be something else than catching the NULL constraint error. Can we abort processing these participants sooner somehow?

PatrickNercessian:
> Adding another problem of the same severity as station_id sounds like a bad idea :P

I think that's just the worst case, in the event that someone intentionally wants to churn participant addresses a lot.

> Hm no I think it would need to be something else than catching the NULL constraint error. Can we abort processing these participants sooner somehow?

Yeah good point. We can just continue the loop if there is no entry in the participant map. But we will still have incomplete data, where we claim to have total measurement counts, but will be missing ones from participants that didn't send any honest measurements in that round

lib/platform-stats.js
```diff
@@ -99,30 +101,136 @@ export const updateDailyStationStats = async (
...
+    const stationStats = statsPerStationParticipant.get(key) ?? { accepted: 0, total: 0 }
+
+    stationStats.total += 1
+    statsPerStationParticipant.set(key, stationStats)
+  }
+
+  debug('Updating recent station details, station_participant_count=%s, participant_subnet_count=%s', statsPerStationParticipant.size, participantSubnetsSet.size)
+
+  // Convert the map and set to arrays for the query
+  const stationParticipants = Array.from(statsPerStationParticipant.keys()).map(k => JSON.parse(k))
```
juliangruber, line 172: can we simplify to `participants`?

juliangruber (changes requested):

index.js
PatrickNercessian: I did! Just modified the other new tests to do two identical calls and assertions

lib/platform-stats.js
PatrickNercessian: Ah I meant to send a message about this. This would presumably also apply to the existing internal function for updateDailyStations, right?

juliangruber: it applies to all nested functions that don't need the parents' scope yeah

lib/platform-stats.js
juliangruber: ping

lib/platform-stats.js
```diff
@@ -99,30 +101,136 @@ export const updateDailyStationStats = async (
...
+ */
+export const updateStationAndParticipantDetails = async (
+  pgClient,
+  honestMeasurements,
+  allMeasurements,
+  participantMap,
+  { day = new Date() } = {}
+) => {
+  const statsPerStationParticipant = new Map()
+  const participantSubnetsSet = new Set()
```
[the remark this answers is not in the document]
PatrickNercessian: Do you mean just renaming, or changing the logic?

If renaming, I think it won't be an accurate name, since the set key is the pair of both together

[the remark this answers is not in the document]
PatrickNercessian: Unless we're going on the assumption that two subnets would not share the same participant, which I guess we do for reward anyway right?

lib/platform-stats.js
```diff
@@ -165,29 +273,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+  const pgClient = await createPgClient()
+
   try {
-    await pRetry(async () => {
-      const pgClient = await createPgClient()
-      try {
-        await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
-      } finally {
-        await pgClient.end()
+    const results = await Promise.allSettled(
```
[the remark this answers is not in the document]
PatrickNercessian: I think we are not doing much sequential logging inside these methods, but that doesn't mean a future PR won't change that, so it's up to you if you'd prefer sequentially calling them instead

lib/platform-stats.js
PatrickNercessian: Ok. Do you concur that it shouldn't happen? A second pair of eyes on that conclusion would be helpful, in case I didn't think about some case or edge case

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: I was suggesting to rename, but you're right that it would be wrong, because of the compound key

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: Ok let's keep this and iterate later, the current approach is simpler

lib/platform-stats.js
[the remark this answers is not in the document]
PatrickNercessian: Ok, I think this also applies to all/most of the other rename suggestions below, right? Each of them have compound keys. What do you think?

index.js
```diff
@@ -3,7 +3,7 @@ import * as Sentry from '@sentry/node'
 import { preprocess } from './lib/preprocess.js'
 import { evaluate } from './lib/evaluate.js'
 import { RoundData } from './lib/round.js'
-import { updateTopMeasurementParticipants } from './lib/platform-stats.js'
+import { periodicDatabaseRefresh } from './lib/platform-stats.js'
```
juliangruber, line 6: can we rename this function to `refreshDatabase`? It shouldn't be concerned with how it's being called (periodic or not)

lib/platform-stats.js
juliangruber: The question is can it happen? If it can, we need to prepare for it. A shouldn't happen isn't safe enough.

Can you think of a case in which it could happen? When would we update 2 days old data?

juliangruber (changes requested):

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: I see! I didn't realize at first that this is a participant x subnet set, and not a set of participant subnets. Tricky to name this! What about a different data structure: Instead of using a compound key for storing each seen participant x subnet combination, we can use a nested data structure, which is clearer:

```js
const subnets = []
subnets[id].push(participant)
```

This way we also don't need to parse the compound key later again.

lib/platform-stats.js
PatrickNercessian: In this case, I was using "should" as synonymous with "can", in the sense that I believe that it can't, but we can never be 100% certain with code, without mathematical, formal verification.

As I think more about it, I am thinking we should do nothing on conflict, and maybe Error or Warning, given that we don't expect it to happen

juliangruber: I agree, in this case it's good to add a log "this shouldn't have happened" instead of conflict resolution logic

juliangruber: > In this case, I was using "should" as synonymous with "can", in the sense that I believe that it can't, but we can never be 100% certain with code, without mathematical, formal verification.

If you believe but don't know then we should handle the case with a log line 👍

lib/platform-stats.js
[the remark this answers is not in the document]
PatrickNercessian: I am not sure I understand. Do you mean:

```
const subnets = []
subnets[id].push(subnet)
```

instead? If so, would this be an array of arrays? And the index of the outermost is the participant_id? But we would not be getting all the participants in each round, so we'd have gaps (if that's even possible in JS, not sure). And how would we use this data structure?

lib/platform-stats.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/e3845c9bd36cfddb719459fe72c3f2389b94ecda

I just removed the ON CONFLICT clause, so that it would throw an error, the DB transaction would rollback, and we get the log output of the catch block.

lib/platform-stats.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/e3845c9bd36cfddb719459fe72c3f2389b94ecda

lib/platform-stats.js
PatrickNercessian: I think this was already addressed (switched ot rsd) before your last comment?

But I also just pushed https://github.com/filecoin-station/spark-evaluate/pull/311/commits/e3845c9bd36cfddb719459fe72c3f2389b94ecda based on a below comment, and spelled out the table name in the "_summary", to align with your suggestion for it for the other summary variable

lib/platform-stats.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/e3845c9bd36cfddb719459fe72c3f2389b94ecda

lib/platform-stats.js
```diff
@@ -11,22 +11,24 @@ const debug = createDebug('spark:platform-stats')
  * @param {import('./preprocess.js').Measurement[]} allMeasurements
  */
 export const updatePlatformStats = async (pgClient, honestMeasurements, allMeasurements) => {
-  const participants = new Set(honestMeasurements.map(m => m.participantAddress))
-  await updateDailyParticipants(pgClient, participants)
+  const participantMap = await mapParticipantsToIds(pgClient, new Set(honestMeasurements.map(m => m.participantAddress)))
```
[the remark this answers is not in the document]
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/90d0c91a3a1000b1b46d9fee6a2074ba77a035a4

index.js
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/e3845c9bd36cfddb719459fe72c3f2389b94ecda

lib/platform-stats.js
```diff
@@ -99,30 +101,136 @@ export const updateDailyStationStats = async (
...
+
+/**
+ * @param {import('pg').Client} pgClient
+ * @param {import('./preprocess.js').Measurement[]} honestMeasurements
+ * @param {import('./preprocess.js').Measurement[]} allMeasurements
+ * @param {Map<string, number>} participantMap
+ * @param {object} options
+ * @param {Date} [options.day]
+ */
+export const updateStationAndParticipantDetails = async (
```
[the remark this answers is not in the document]
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/9757960df2a165e6f7acf5cc2801d2ee7afe6948

PatrickNercessian:
ping @juliangruber on https://github.com/filecoin-station/spark-evaluate/pull/311#issuecomment-2284784928

lib/platform-stats.js
juliangruber: what does rsd stand for?

index.js
juliangruber: you renamed it to `databaseRefresh`, can we do the more intuitive `refreshDatabase`? To me functions should be verbs, as they're doing something

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: Hm no I meant it the way I wrote it. We need to keep track of stats per subnet per participant. So for each subnet, you would keep track of the participants that are in that subnet. Thinking about it more, an object of objects could work better for later accessing the data:

```js
const subnets = {}
subnets[subnetId][participantId] = { total: 0, accepted: 0 }
```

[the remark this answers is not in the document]
juliangruber: or am I misunderstanding this part?

juliangruber:
Thanks for the ping!

> > Adding another problem of the same severity as station_id sounds like a bad idea :P
> 
> I think that's just the worst case, in the event that someone intentionally wants to churn participant addresses a lot, _and_ for whatever reason they're often marked as all non-honest measurements forever (which is the only case where they'd be included now but not before this change, because if they _ever_ submit an honest measurement, then they'd be added to the database regardless.
> 

I don't see anyone wanting to churn this table, or even knowing about it really. I think we only need to worry about the natural use we're already seeing.

> > Hm no I think it would need to be something else than catching the NULL constraint error. Can we abort processing these participants sooner somehow?
> 
> Yeah good point. We can just continue the loop if there is no entry in the participant map. But we will still have incomplete data, where we claim to have total measurement counts, but will be missing ones from participants that didn't send any honest measurements in that round

I can't make a good judgement on this one. Is it bad in this case to have a lower total count? If we need this total count to include unaccepted measurements, then let's collect all the data instead.

PatrickNercessian:
> I think we only need to worry about the natural use we're already seeing.

I think I agree. So should we go ahead and process all measurements into `participants` table using `mapParticipantsToIds()`?

> I can't make a good judgement on this one. Is it bad in this case to have a lower total count? If we need this total count to include unaccepted measurements, then let's collect all the data instead.

I have a personal aversion toward incomplete data (especially when it's not clear what data is missing). I think there's not a direct issue with missing this total measurement count data today, but maybe later analysis will be more cumbersome or nuanced on that data if we don't have it.

PatrickNercessian:
Ping @juliangruber (and @bajtos if back in office) on the above discussion

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: ping @PatrickNercessian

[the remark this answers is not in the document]
PatrickNercessian: Oops, sorry, I never noticed this response. So we have the two data structures for separate reasons:

```
  const statsPerStationParticipant = new Map()
  const participantSubnetsSet = new Set()
```

1. `statsPerStationParticipant` uses a compound key of `stationId` and `participantId`, and tracks the stats for that pair. It's used to insert into `recent_station_details` (because those columns, along with `day`, are the compound primary key of that table).
2. `participantSubnetsSet` uses a compound key of `participantId` and `subnet`. It's used to insert into `recent_participant_subnets` (because those columns, along with `day`, are the compound primary key of that table)

However, we could potentially use a nested structure for `statsPerStationParticipant`, similar to what you described (but the nested keys being `stationId` and `participantId` instead). But we still need the functionality allowed by `participantSubnetsSet`, and I don't see a good way of matching that structure (aside from another nested object structure, where the value is something arbitrary that we never use).

Also, the reason the code has evolved to these compound string keys instead is that we already do this for the existing `daily_stations` code, which uses a triple compound key. Would we want to also turn this into a 3x nested structure? I guess we are eventually planning to remove this `daily_stations` table and code, so maybe we can ignore it for now?

juliangruber:
## How participants are defined

You count as a participant even if your measurement was rejected, because say with the upcoming committees logic (or the subnet deduplication part) an honest participant can have rejected measurements, and if they have bad luck they will have _only_ rejected measurements.

1. In a separate PR: Change updateDailyParticipants to use _all_ measurements
1. Follow up and do the same in this PR

## What happens on high participant churn?

_TBD_

lib/platform-stats.js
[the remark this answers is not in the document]
juliangruber: @bajtos and me discussed this and came to the following conclusion:

1. It is cleaner to use traditional nested data structures. There's a high maintenance cost when introducing a novel data structure for a single file.
2. For performance, let's use `Map` instead of `{}`

Example of how the nested data structure works:

```js
const subnets = new Map()

  if (!subnets.has(m.inet_group)) {
    subnets.set(m.inet_group, new Set())
  }
  subnets.get(m.inet_group).add(participantId)
}

await pgClient.query(`
  INSERT INTO recent_participant_subnets (day, participant_id, subnet)
  SELECT $1::DATE, unnest($2::int[]), unnest($3::text[])
  ON CONFLICT (day, participant_id, subnet) DO NOTHING
, [
  day,
  [...subnets.entries().flatMap(([subnet, participantIds]) => participantIds.map(() => subnet))],
  [...subnets.entries().flatMap(([_, participantIds]) => participantIds)]
])
```

juliangruber (changes requested):
see
- https://github.com/filecoin-station/spark-evaluate/pull/311#issuecomment-2309758205
- https://github.com/filecoin-station/spark-evaluate/pull/311#discussion_r1731087371

lib/platform-stats.js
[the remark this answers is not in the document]
PatrickNercessian: Addressed in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/cee07ab2dc0d95c234ee32d2bd6e1fd80072d72a and https://github.com/filecoin-station/spark-evaluate/pull/311/commits/3bf00da3d431da64b0a5d84e00bdeda05ac04c7d

lib/platform-stats.js
```diff
@@ -99,30 +93,153 @@ export const updateDailyStationStats = async (
...
+    subnets.set(participantId, participantSubnets) // in case new set was created
+  }
+
+  for (const m of allMeasurements) {
+    if (m.stationId == null) continue
+    const participantId = participantsMap.get(m.participantAddress)
+    if (participantId == null) continue
+
+    updateParticipantsStationsStats(m, participantId, 'total')
+  }
```
PatrickNercessian, line 167: Do we want to also add to the participantSubnets for allMeasurements?

```
    const participantSubnets = subnets.get(participantId) ?? new Set()
    participantSubnets.add(m.inet_group)
    subnets.set(participantId, participantSubnets) // in case new set was created
```

We were already not doing this, but I don't recall if there was a specific reason.

bajtos: > Do we want to also add to the participantSubnets for allMeasurements?

Yes. We should look at `allMeasurements` when counting anything related to participants - participant addresses, IP subnets, and station IDs too.

Could you please check if we collect station IDs from all measurements, not only the accepted ones?

PatrickNercessian: Gotcha, adding that in now.

Can you elaborate on what you mean by "collect station IDs"? We call `updateParticipantsStationsStats` in the code attached to this comment thread for allMeasurements to increment the total, is that what you mean?

lib/platform-stats.js
```diff
@@ -11,38 +11,32 @@ const debug = createDebug('spark:platform-stats')
...
   await updateDailyStationStats(pgClient, honestMeasurements, allMeasurements)
+  await updateStationsAndParticipants(pgClient, honestMeasurements, allMeasurements, participantsMap)
 }
 
 /**
  * @param {import('pg').Client} pgClient
  * @param {import('./preprocess.js').Measurement[]} honestMeasurements
  * @param {import('./preprocess.js').Measurement[]} allMeasurements
- * @param {Date} day
+ * @param {object} options
```
juliangruber, line 24: This argument is also optional

```suggestion
 * @param {object} [options]
```

lib/platform-stats.js
```diff
@@ -165,29 +284,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
   `, [
     newAddresses
   ])
 
   assert.strictEqual(created.length, newAddresses.length)
-  return ids.concat(created.map(r => r.id))
+  // eslint-disable-next-line camelcase
+  for (const { id, participant_address } of created) {
+    participantsMap.set(participant_address, id)
+  }
```
juliangruber, line 298:
```suggestion
  for (const { id, participant_address: participantAddress } of created) {
    participantsMap.set(participantAddress, id)
  }
```

lib/platform-stats.js
```diff
@@ -165,29 +284,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+
+  return participantsMap
 }
 
 /**
  * @param {import('./typings.js').CreatePgClient} createPgClient
+ * @param {object} options
```
juliangruber, line 305:
```suggestion
 * @param {object} [options]
```

This argument is also optional

lib/platform-stats.js
```diff
@@ -165,29 +284,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+ * @param {Array<Function>} [options.functionsToRun]
+ */
+export const refreshDatabase = async (
+  createPgClient,
+  { functionsToRun = [updateTopMeasurementParticipants, aggregateAndCleanupRecentData, updateMonthlyActiveStationCount] } = {}
```
juliangruber, line 310:
```suggestion
  {
    functionsToRun = [
      updateTopMeasurementParticipants,
      aggregateAndCleanupRecentData,
      updateMonthlyActiveStationCount
    ]
  } = {}
```

just to make it a bit easier to read

lib/platform-stats.js
```diff
@@ -165,29 +284,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+  const pgClient = await createPgClient()
+
+  try {
+    const results = await Promise.allSettled(
+      functionsToRun.map(func => func(pgClient))
+    )
+
+    results.forEach((result, index) => {
+      if (result.status === 'rejected') {
+        console.error(`Error running function ${functionsToRun[index].name}:`, result.reason)
```
juliangruber, line 321:
```suggestion
        console.error(`Error running function ${functionsToRun[index].name}:`, result.reason)
        Sentry.captureException(result.reason)
```

lib/platform-stats.js
```diff
@@ -165,29 +284,55 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+      }
+    })
+  } finally {
+    await pgClient.end()
+  }
```
juliangruber, line 326:
```suggestion
  const results = await Promise.allSettled(
    functionsToRun.map(func => func(pgClient))
  )

  results.forEach((result, index) => {
    if (result.status === 'rejected') {
      console.error(`Error running function ${functionsToRun[index].name}:`, result.reason)
    }
  })

  await pgClient.end()
```

No `try/catch` necessary, as `Promise.allSettled` doesn't reject

lib/platform-stats.js
```diff
@@ -199,3 +344,121 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
     Sentry.captureException(err)
   }
 }
+
+/**
+ * @param {import('pg').Client} pgClient
+ */
+export const aggregateAndCleanupRecentData = async (pgClient) => {
```
juliangruber, line 351:
```suggestion
export const aggregateAndCleanUpRecentData = async (pgClient) => {
```

"to cleanup" doesn't exist :)

juliangruber: Also needs to be renamed above

lib/platform-stats.js
```diff
@@ -199,3 +344,121 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+    await pgClient.query(`
+      DELETE FROM recent_participant_subnets WHERE day <= CURRENT_DATE - INTERVAL '2 days'
+    `)
+
+    await pgClient.query('COMMIT')
+  } catch (err) {
+    console.error('Error aggregating and cleaning up recent data', err)
+    await pgClient.query('ROLLBACK')
+    Sentry.captureException(err)
+  }
```
juliangruber, line 417:
```suggestion
  } catch (err) {
    await pgClient.query('ROLLBACK')
    throw errr
  }
```

This lets the outer function handle the error and submit it to Sentry

lib/platform-stats.js
```diff
@@ -199,3 +344,121 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+          AND day < DATE_TRUNC('month', CURRENT_DATE);
+      `)
+
+      await pgClient.query('COMMIT')
+    }
+  } catch (err) {
+    console.error('Error updating monthly active station count', err)
+    await pgClient.query('ROLLBACK')
+    Sentry.captureException(err)
+  }
```
juliangruber, line 455:
```suggestion
  } catch (err) {
    await pgClient.query('ROLLBACK')
    throw err
  }
```

PatrickNercessian:
> Just minor final change requests 👍 
> 
> 
> 
> I believe this can be landed later today, @PatrickNercessian is this PR still a draft on purpose?

Yes, converted to a draft because it had a similar change as #328, and if that landed first (it did) we need to resolve the merge conflict. Maybe that's not worthy of it being a draft, but I wasn't sure.

Will update according to the merge conflict and your comments this morning

juliangruber:
> Yes, converted to a draft because it had a similar change as https://github.com/filecoin-station/spark-evaluate/pull/328, and if that landed first (it did) we need to resolve the merge conflict. Maybe that's not worthy of it being a draft, but I wasn't sure.

Understood, no need to mark it as a draft unless it shouldn't be landed before the other PR

juliangruber (approved):
I've pinged @bajtos on Slack asking for final review. For me this is good to go!

PatrickNercessian:
Sorry, forgot to remove the comment about ON CONFLICT during [this commit](https://github.com/PatrickNercessian/spark-evaluate/commit/e3845c9bd36cfddb719459fe72c3f2389b94ecda). Just did so in https://github.com/filecoin-station/spark-evaluate/pull/311/commits/37b972ba4d82634cd0a08067ba0eb8c2579fd0f6

lib/platform-stats.js
bajtos: I was wondering whether the code inserting data to `daily_stations` would include Station IDs whose measurements we rejected (Station IDs with no accepted measurements).

I checked the source code of `updateDailyStationStats` and verified that we include those Stations IDs.

All is good.

https://github.com/filecoin-station/spark-evaluate/blob/ae885a9dcac95a9c3265634511eae5779e7370a5/lib/platform-stats.js#L25C14-L98

lib/platform-stats.js
```diff
@@ -199,3 +346,117 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+ * @param {import('pg').Client} pgClient
+ */
+export const updateMonthlyActiveStationCount = async (pgClient) => {
+  try {
+    const result = await pgClient.query(`
+      SELECT 1 FROM monthly_active_station_count
+      WHERE month = DATE_TRUNC('month', CURRENT_DATE - INTERVAL '1 month')
+    `)
+
+    if (result.rowCount === 0) {
```
bajtos, line 429: Let's use reverse logic + early return to make this code easier to read and avoid extra nesting.

```js
if (result.rowCount === 0) return
```

bajtos (changes requested):
Great work, @PatrickNercessian!

I apologise for sending my review a bit late.

I reviewed the implementation and DB schema migration script. I'll review the tests later.

lib/platform-stats.js
```diff
@@ -11,38 +11,32 @@ const debug = createDebug('spark:platform-stats')
...
-    stationId: m.stationId,
-    participantAddress: m.participantAddress,
-    inet_group: m.inet_group
-  })
-
   for (const m of honestMeasurements) {
     if (m.stationId == null) continue
 
-    const key = getKey(m)
+    const key = getStatsKey(m)
```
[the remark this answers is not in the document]
juliangruber: > When getKey is defined inside updateDailyStationStats, then its purpose is clear - it calculates the key we use to group measurements for calculating "daily station stats".
>
> When you move it outside of updateDailyStationStats, it's no longer clear what key it computes.

Especially since this will be removed anyway, I don't want to bikeshed about this part. I'm thinking:

- Moving the new function close to this function helps with discoverability. I also don't like when the new function is so far away
- Reverting that part and inline also works, to keep this diff small

Please whoever feels strongly just make a decision and let's finish up this work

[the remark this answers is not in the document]
PatrickNercessian: Will stick with inline, since the change is now outside the scope of this PR

lib/platform-stats.js
```diff
@@ -99,30 +93,155 @@ export const updateDailyStationStats = async (
...
+    updateParticipantsStationsStats(m, participantId, 'total')
+
+    let subnetsSet = subnets.get(participantId)
+    if (!subnetsSet) {
+      subnetsSet = new Set()
+      subnets.set(participantId, subnetsSet)
+    }
+
+    subnetsSet.add(m.inet_group)
+  }
```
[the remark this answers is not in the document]
PatrickNercessian: Sounds good! The only downside I see is that we are spreading apart the logic that decides which measurements are accepted vs not, rather than having it in one single place. But it would make this simpler code, and we already do it elsewhere, so I'll make the change

lib/platform-stats.js
```diff
@@ -165,29 +284,57 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
+  const results = await Promise.allSettled(
+    functionsToRun.map(func => func(pgClient))
+  )
+  results.forEach((result, index) => {
+    if (result.status === 'rejected') {
+      console.error(`Error running function ${functionsToRun[index].name}:`, result.reason)
+      Sentry.captureException(result.reason)
+    }
+  })
+  await pgClient.end()
```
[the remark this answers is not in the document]
PatrickNercessian: Sure, we can do them sequentially since the total time shouldn't matter too much, and avoid any fears of slowing down other writes.

lib/platform-stats.js
```diff
@@ -165,29 +284,57 @@ export const mapParticipantsToIds = async (pgClient, participantsSet) => {
...
     await pRetry(async () => {
-      const pgClient = await createPgClient()
-      try {
-        await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
-      } finally {
-        await pgClient.end()
-      }
+      await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
     }, {
       retries: 3,
```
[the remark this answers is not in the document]
PatrickNercessian: I remember the context was because `index.js` [calls this function today directly](https://github.com/filecoin-station/spark-evaluate/blob/ae885a9dcac95a9c3265634511eae5779e7370a5/index.js#L144), and only has access to the `createPgClient` function (i.e. it does not instantiate a pgClient). With these changes, since we are now doing more than just `updateTopMeasurementParticipants` during the scheduled execution, I "bubbled up" the creation of the client to be used for each function.

We could instead pass `createPgClient` to each of the `functionsToRun` in `refreshDatabase()`, and have them each responsible for creating a new client. What would you prefer?

lib/platform-stats.js
```diff
@@ -199,3 +346,117 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
     Sentry.captureException(err)
   }
 }
+
+/**
+ * @param {import('pg').Client} pgClient
+ */
+export const aggregateAndCleanUpRecentData = async (pgClient) => {
+  try {
+    await pgClient.query('BEGIN')
```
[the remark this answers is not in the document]
PatrickNercessian: Yes, as I understand it.

lib/platform-stats.js
```diff
@@ -99,30 +93,155 @@ export const updateDailyStationStats = async (
...
+  ])
+
+  await pgClient.query(`
+    INSERT INTO recent_participant_subnets (day, participant_id, subnet)
+    SELECT $1::DATE, unnest($2::int[]), unnest($3::text[])
+    ON CONFLICT (day, participant_id, subnet) DO NOTHING
+  `, [
+    day,
+    participantsSubnets.map(ps => ps.participantId),
+    participantsSubnets.map(ps => ps.subnet)
```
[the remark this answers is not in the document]
PatrickNercessian: Yeah I think that makes sense!

lib/platform-stats.js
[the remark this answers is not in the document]
bajtos: > I remember the context was because `index.js` [calls this function today directly](https://github.com/filecoin-station/spark-evaluate/blob/ae885a9dcac95a9c3265634511eae5779e7370a5/index.js#L144), and only has access to the `createPgClient` function (i.e. it does not instantiate a pgClient). > 

Thanks, that makes sense!

> With these changes, since we are now doing more than just `updateTopMeasurementParticipants` during the scheduled execution, I "bubbled up" the creation of the client to be used for each function.

Makes sense 👍🏻 

> We could instead pass `createPgClient` to each of the `functionsToRun` in `refreshDatabase()`, and have them each responsible for creating a new client. What would you prefer?

If we pass `createPgClient` to each function to run, then each function must deal with the complexity of creating a new client and releasing it in a `finally` block. I think this additional complexity is not worth it right now.

Let's keep what you have now in this pull request. 👍🏻

test/platform-stats.test.js
```diff
@@ -226,6 +235,191 @@ describe('platform-stats', () => {
...
+          participant_id,
+          accepted_measurement_count,
+          total_measurement_count
+        FROM recent_station_details
+        WHERE day = $1::DATE
+        ORDER BY station_id, participant_id
+      `, [today])
+
+      assert.strictEqual(stationDetails.length, 2)
+      assert.deepStrictEqual(stationDetails, [
```
bajtos, line 297: `deepStrictEqual` will implicitly assert the array size and it will provide a more descriptive error message. I propose removing the first assertion.

```suggestion
      assert.deepStrictEqual(stationDetails, [
```

bajtos (commented):
The tests looks pretty good, I have only one comment to consider.

test/platform-stats.test.js
bajtos: (The same comment applies to other places asserting array length first and array contents second.)

bajtos (approved):
The changes LGTM, as far as I am concerned.

@juliangruber would you like to take another look at the recent changes in `refreshDatabase` implementation?

bajtos:
> The changes LGTM, as far as I am concerned.
> 
> @juliangruber would you like to take another look at the recent changes in `refreshDatabase` implementation?

See https://github.com/filecoin-station/spark-evaluate/pull/311#discussion_r1735908050

Another thread you may be interested in: https://github.com/filecoin-station/spark-evaluate/pull/311#discussion_r1735913367

bajtos (approved):
👌🏻

juliangruber (approved):
@bajtos thank you for pointing me at these relevant places! All is looking good to me :)

I propose to deploy this tomorrow morning

bajtos:
Landed! 🚀

lib/platform-stats.js
```diff
@@ -199,3 +349,109 @@ export const updateTopMeasurementParticipants = async (createPgClient) => {
...
+        SELECT
+          day,
+          SUM(accepted_measurement_count) AS accepted_measurement_count,
+          SUM(total_measurement_count) AS total_measurement_count,
+          COUNT(DISTINCT station_id) AS station_count,
+          COUNT(DISTINCT participant_id) AS participant_address_count
+        FROM
+          recent_station_details
+        WHERE
+          day <= CURRENT_DATE - INTERVAL '2 days'
```
bajtos, line 371: @PatrickNercessian @juliangruber do you happen to remember why we decided to use `INTERVAL '2 days'` here?

The consequence is that we don't have data for yesterday, e.g. to populate the "daily stations" chart.

If the intention was to ensure that `recent_station_details_summary` has complete data for the day being processed, then I think `INTERVAL '26 hours'` should be good enough.

Thoughts?

juliangruber: I don't know tbh

PR #320, opened 2024-08-12
fix don't retry scoring wrong round

juliangruber:
https://my.papertrailapp.com/systems/9b22/events?focus=1759713909884985352&selected=1759713907217408007


bajtos (commented):
Awesome! Thank you for adding tests for both paths.

Issue #312, opened 2024-08-09
Outage because of "JsonRpcProvider failed to detect network and cannot start up; retry in 1s"


bajtos:
While investigating https://github.com/filecoin-station/core/issues/569, I noticed that spark-evaluate logs are full of the following error messages:

```
2024-08-09T07:19:31Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:21:42Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:23:53Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:26:04Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:28:15Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:30:26Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:32:37Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:34:48Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:36:59Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
2024-08-09T07:39:10Z app[e2867541be3e68] cdg [info]JsonRpcProvider failed to detect network and cannot start up; retry in 1s (perhaps the URL is wrong or the node is not started)
[...]

Issue #281, opened 2024-07-03
Use FallbackProvider to improve reliability


juliangruber:
Similar to other services, this one can use ether's FallbackProvider to automatically switch to different JSON-RPC providers, when one behaves badly.

PR #325, opened 2024-08-22
Add exit when rpc backend unavailable

juliangruber:
It looks like spark-evaluate isn't properly rotating rpc-backends and only logging errors https://my.papertrailapp.com/events?q=spark-evaluate


lib/ie-contract.js
```diff
@@ -22,5 +23,16 @@ export const createMeridianContract = async (contractAddress = SparkImpactEvalua
     SparkImpactEvaluator.ABI,
     provider
   )
+
+  try {
+    await pTimeout(ieContract.currentRoundIndex(), {
+      milliseconds: 10_000
```
juliangruber, line 29: should I increase this to 20s or 60s

juliangruber: Don't want to give up too soon in case a backend is just slow but overall available

lib/ie-contract.js
```diff
@@ -22,5 +23,16 @@ export const createMeridianContract = async (contractAddress = SparkImpactEvalua
...
+    })
+  } catch (err) {
+    console.error(`Failed to connect to contract via ${RPC_URL}`)
+    console.error(err)
+    process.exit(1)
+  }
```
bajtos, line 35: Should we send this error to Sentry?

Also, why `process.exit`? Isn't it enough to (re)throw the error and let the "unhandled rejection" handler exit the process? That way, the error should be automatically captured & sent to Sentry.

lib/ie-contract.js
juliangruber: another alternative: Start using FallbackProvider. Wdyt @bajtos?

bajtos (approved):
LGTM as a quick fix to solve the immediate problem.

For the longer term, I'd like us to consider the questions below.

lib/ie-contract.js
juliangruber: Good idea! I hadn't considered just throwing

lib/ie-contract.js
bajtos: > This surfaces another issue: What if the backend becomes unavailable later on? Would be better to make the process fail then as well

💯 

I think it will be more often that the RPC backend becomes unavailable while the service is running, rather than at the service startup.

> another alternative: Start using FallbackProvider. Wdyt @bajtos?

IIRC, we have had a good experience with FallbackProvider in one of the other Spark services. I think it's a great idea to use it in spark-evaluate too.

bajtos: > should I increase this to 20s or 60s?
>
> Don't want to give up too soon in case a backend is just slow but overall available

I think 10 seconds is a good start; we can easily tweak this value later based on what we learn about real-world performance.

Have you considered what happens when none of the providers can respond in 10 seconds? Our service will always exit after ~10 seconds; Fly.io will restart it then we exit again, etc.

I think it would be great to create a Papertrail->Slack alert when this happens for more than N minutes, where N can be like 5-10 minutes.

@juliangruber WDYT?

juliangruber: > Have you considered what happens when none of the providers can respond in 10 seconds? Our service will always exit after ~10 seconds; Fly.io will restart it then we exit again, etc.

This papertrail search should handle it:

> "spark-evaluate errors" search found 5 matches in the past ten minutes 

Do you think it's worth the extra effort?

juliangruber: Related: https://github.com/filecoin-station/spark-evaluate/issues/312

juliangruber: I'm adding https://github.com/filecoin-station/spark-evaluate/issues/281 to the current milestone

bajtos: > > Have you considered what happens when none of the providers can respond in 10 seconds? Our service will always exit after ~10 seconds; Fly.io will restart it then we exit again, etc.
> 
> This papertrail search should handle it:
> 
> > "spark-evaluate errors" search found 5 matches in the past ten minutes
> 
> Do you think it's worth the extra effort?

I think we want an alert to let us know when this is happening, don't we? Is there any other alert that would be eventually triggered when spark-evaluate keeps crashing at startup? (E.g. no round evaluation scores submitted or no new RSR entries created.) If yes, then that's good enough for me.

bajtos:
> ## Suspect Issues
> 
> * ‼️ **TimeoutError: Promise timed out after 10000 milliseconds** `createMeridianContract(ie-contract)` [View Issue](https://space-meridian.sentry.io/issues/5757988954/?referrer=github-pr-bot)
> 

@juliangruber could you please take a look?

lib/ie-contract.js
juliangruber: Well, isn't that the alert that already exists?

juliangruber:
https://github.com/filecoin-station/spark-evaluate/pull/333

PR #329, opened 2024-08-26
Add auto cancel stuck transactions. Closes #315

### LINKED ISSUE #315: Cancel pending transactions that are too old
When the gas base fee suddenly and significantly increases, our estimation of the required gas amount is too low. No miner accepts the message, and it sits in the queue until the gas base fee decreases again. It usually takes several hours before the base fee decreases and our transactions are committed again.

When this happens, some transactions are committed with hours-long delays. Our smart contract rejects such transactions because evaluation scores can be reported only for the previous round, and the rounds advanced many times while the message was stuck waiting to be added to the chain.

We should implement an automated mechanism that detects transactions meeting the following criteria:
- The transaction is still pending (in the mempool, not added to any block).
- The transaction was created before the previous round started.

Then we should proceed to cancel these transactions automatically.

The goal is to avoid the situation depicted in the screenshot below, where we have >400 pending transactions that will be later rejected by the smart contract.

Screenshot 2024-08-09 at 12 06 18

juliangruber:
Closes #315

On an interval, check the mempool for stuck transactions, and cancel them all.

In this first simple implementation, a transaction is considered "stuck" when it's older than the current IE round length of 20 minutes. This can later be improved by parsing the transaction payload and comparing the round index with the current IE round index.

Example output:

```
Checking for stuck transactions...
Transactions to cancel:
- bafy2bzaceborkhyvzoq3gfuqqy5z7q647yx2agyy6l6zshz3lhviaskf6cdew (age 20m)
- bafy2bzaceajr6r24yefuonrmbpo4rs756xrrud2knl6z7towqemomtxq7uv34 (age 20m)
- bafy2bzacecqq7o6mdvbddm7gxyfcsuo36lidcva32ajwicyfm4yjymrhhvwye (age 20m)
- bafy2bzaceaczqfxhcaqv4dhq2nqdfa5yhpn3gx27q337ntyrk76cghi4dd4ny (age 20m)
Calculating gas fees from the recent Send message bafy2bzacecspzxpfjfw4gdusl2xulivy3w6lk3qxvgpww5abp5kgh34s3qyni (created at 2024-08-27T07:16:00.000Z)
Replacing bafy2bzaceborkhyvzoq3gfuqqy5z7q647yx2agyy6l6zshz3lhviaskf6cdew...
Replacing bafy2bzaceajr6r24yefuonrmbpo4rs756xrrud2knl6z7towqemomtxq7uv34...
Replacing bafy2bzacecqq7o6mdvbddm7gxyfcsuo36lidcva32ajwicyfm4yjymrhhvwye...
Replacing bafy2bzaceaczqfxhcaqv4dhq2nqdfa5yhpn3gx27q337ntyrk76cghi4dd4ny...
Waiting for receipt of replacement 0xb4a42acb05c7c3048351776109669deb8bfdc9ce11117ba4b83ac9caa30ad41b for bafy2bzaceborkhyvzoq3gfuqqy5z7q647yx2agyy6l6zshz3lhviaskf6cdew
Waiting for receipt of replacement 0xe39b4b1bf917d03fb6f53fe39b9b03899a02a887bfbcaf6acf6c4cc49d0988b8 for bafy2bzaceajr6r24yefuonrmbpo4rs756xrrud2knl6z7towqemomtxq7uv34
Waiting for receipt of replacement 0xfc9994b2fbcfd5b8c656dde890ca6d04a953d3d6dd8b6ed86420e8cb235acebe for bafy2bzacecqq7o6mdvbddm7gxyfcsuo36lidcva32ajwicyfm4yjymrhhvwye
Waiting for receipt of replacement 0x29c6721a15576c7a9cb8829968dadb0b90778ea4b38384db385b23b09d07f4d9 for bafy2bzaceaczqfxhcaqv4dhq2nqdfa5yhpn3gx27q337ntyrk76cghi4dd4ny
Replaced bafy2bzaceborkhyvzoq3gfuqqy5z7q647yx2agyy6l6zshz3lhviaskf6cdew
Replaced bafy2bzaceajr6r24yefuonrmbpo4rs756xrrud2knl6z7towqemomtxq7uv34
Replaced bafy2bzacecqq7o6mdvbddm7gxyfcsuo36lidcva32ajwicyfm4yjymrhhvwye
Replaced bafy2bzaceaczqfxhcaqv4dhq2nqdfa5yhpn3gx27q337ntyrk76cghi4dd4ny
```

### DIFF
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -91,7 +91,7 @@ export async function getRecentSendMessage () {
   return /** @type {any} */(await res.json())
 }
 
-const cancelTx = async ({ tx, signer, gasUsed, gasFeeCap }) => {
+const cancelTx = async ({ tx, signer, recentGasUsed, recentGasFeeCap }) => {
   const oldGasPremium = Number(tx.gasPremium)
   const nonce = tx.nonce
 
@@ -101,8 +101,8 @@ const cancelTx = async ({ tx, signer, gasUsed, gasFeeCap }) => {
       to: signer.address,
       value: 0,
       nonce,
-      gasLimit: Math.ceil(gasUsed * 1.1),
-      maxFeePerGas: gasFeeCap,
+      gasLimit: Math.ceil(recentGasUsed * 1.1),
+      maxFeePerGas: recentGasFeeCap,
       // Increase by 25% + 1 attoFIL (easier: 25.2%)
       maxPriorityFeePerGas: Math.ceil(oldGasPremium * 1.252)
     })
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -46,30 +46,30 @@ export async function getMessagesInMempool (f4addr) {
 
 /**
  * @returns {Promise<{
-  "cid": string;
-  "height": number;
-  "timestamp": number;
-  "gasLimit": number;
-  "gasFeeCap": string;
-  "gasPremium": string;
-  "method": string;
-  "methodNumber": number;
-  "receipt": {
-      "exitCode": number;
-      "return": string;
-      "gasUsed": number;
-    },
-    "size": number;
-    "error": string;
-    "baseFee": string;
-    "fee": {
-      "baseFeeBurn": string;
-      "overEstimationBurn": string;
-      "minerPenalty": string;
-      "minerTip": string;
-      "refund": string;
-    },
-  }>}
+  cid: string;
+  height: number;
+  timestamp: number;
+  gasLimit: number;
+  gasFeeCap: string;
+  gasPremium: string;
+  method: string;
+  methodNumber: number;
+  receipt: {
+    exitCode: number;
+    return: string;
+    gasUsed: number;
+  },
+  size: number;
+  error: string;
+  baseFee: string;
+  fee: {
+    baseFeeBurn: string;
+    overEstimationBurn: string;
+    minerPenalty: string;
+    minerTip: string;
+    refund: string;
+  },
+}>}
 */
 export async function getRecentSendMessage () {
   let res = await fetch('https://filfox.info/api/v1/message/list?method=Send')
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -3,7 +3,7 @@ import assert from 'node:assert'
 import * as Sentry from '@sentry/node'
 import ms from 'ms'
 
-const ROUND_LENGTH_MS = 20 * 60 * 1000 // 20 minutes
+const ROUND_LENGTH_MS = ms('20 minutes')
 
 /**
   * @param {string} f4addr
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -144,11 +144,11 @@ const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
     new Date(recentSendMessage.timestamp * 1000).toISOString()
   )
 
-  const gasUsed = recentSendMessage.receipt.gasUsed
-  const gasFeeCap = Number(recentSendMessage.gasFeeCap)
+  const recentGasUsed = recentSendMessage.receipt.gasUsed
+  const recentGasFeeCap = Number(recentSendMessage.gasFeeCap)
 
   await Promise.all(txsToCancel.map(async tx => {
-    await cancelTx({ tx, signer, gasUsed, gasFeeCap })
+    await cancelTx({ tx, signer, recentGasUsed, recentGasFeeCap })
   }))
 
   return true
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -95,7 +95,7 @@ export const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
   const messages = await getMessagesInMempool(walletDelegatedAddress)
   const txsToCancel = messages.filter(m => {
     return new Date(m.createTimestamp * 1000) <
-      new Date(Date.now() - ROUND_LENGTH_MS)
+      new Date(Date.now() - (2 * ROUND_LENGTH_MS))
   })
   if (txsToCancel.length === 0) {
     console.log('No transactions to cancel')
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -168,7 +168,7 @@ export const startCancelStuckTxs = async ({
       Sentry.captureException(err)
     }
     if (!didCancelTxs) {
-      await timers.setTimeout(60_000)
+      await timers.setTimeout(ms('1 minute'))
     }
   }
 }
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -108,7 +108,7 @@ const cancelTx = async ({ tx, signer, gasUsed, gasFeeCap }) => {
       `Waiting for receipt of replacement ${replacementTx.hash} for ${tx.cid}`
     )
     await replacementTx.wait()
-    console.log(`Replaced ${tx.cid}`)
+    console.log(`Replaced ${tx.cid} with ${replacementTx.hash}`)
   } catch (err) {
     console.error(err)
     Sentry.captureException(err)
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -2,6 +2,7 @@ import pRetry from 'p-retry'
 import assert from 'node:assert'
 import * as Sentry from '@sentry/node'
 import ms from 'ms'
+import timers from 'node:timers/promises'
 
 const ROUND_LENGTH_MS = ms('20 minutes')
 
@@ -89,7 +90,7 @@ async function getRecentSendMessage () {
   return /** @type {any} */(await res.json())
 }
 
-export const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
+const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
   console.log('Checking for stuck transactions...')
 
   const messages = await getMessagesInMempool(walletDelegatedAddress)
@@ -146,4 +147,25 @@ export const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
   }))
 
   return true
+}
+
+export const startCancelStuckTxs = async ({
+  walletDelegatedAddress,
+  signer
+}) => {
+  while (true) {
+    let didCancelTxs = false
+    try {
+      didCancelTxs = await cancelStuckTxs({
+        walletDelegatedAddress,
+        signer
+      })
+    } catch (err) {
+      console.error(err)
+      Sentry.captureException(err)
+    }
+    if (!didCancelTxs) {
+      await timers.setTimeout(60_000)
+    }
+  }
 }
--- a/lib/cancel-stuck-txs.js
+++ b/lib/cancel-stuck-txs.js
@@ -1,0 +1,144 @@
+import pRetry from 'p-retry'
+import assert from 'node:assert'
+import * as Sentry from '@sentry/node'
+
+const ROUND_LENGTH_MS = 20 * 60 * 1000 // 20 minutes
+
+/**
+   * @param {string} f4addr
+   * @returns 1000 oldest messages
+   */
+async function getMessagesInMempool (f4addr) {
+  const res = await pRetry(
+    () => fetch(`https://filfox.info/api/v1/message/mempool/filtered-list?address=${f4addr}&pageSize=1000`),
+    {
+      async onFailedAttempt (error) {
+        console.warn(error)
+        console.warn('Filfox request failed. Retrying...')
+      }
+    }
+  )
+  if (!res.ok) {
+    throw new Error(`Filfox request failed with ${res.status}: ${(await res.text()).trimEnd()}`)
+  }
+  /** @type {{
+     messages: {
+      cid: string;
+      from: string;
+      to: string;
+      nonce: number;
+      value: string;
+      gasLimit: number;
+      gasFeeCap: string;
+      gasPremium: string;
+      method: string;
+      methodNumber: number;
+      evmMethod: string;
+      createTimestamp: number;
+    }[];
+  }}
+  */
+  const { messages } = /** @type {any} */(await res.json())
+  return messages
+}
+
+/**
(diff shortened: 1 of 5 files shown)


lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,144 @@
...
+  const recentSendMessage = await getRecentSendMessage()
+  console.log('Calculating gas fees from the recent Send message %s (created at %s)',
+    recentSendMessage.cid,
+    new Date(recentSendMessage.timestamp * 1000).toISOString()
+  )
+
+  const gasUsed = recentSendMessage.receipt.gasUsed
+  const gasFeeCap = Number(recentSendMessage.gasFeeCap)
+
+  await Promise.all(txsToCancel.map(async tx => {
```
juliangruber, line 118: Try to cancel as many as possible in parallel, therefore await transaction confirmation. If we went serial (with a recently measured cancel duration of 13 minutes), we would likely never catch up with the queue. And if we didn't await confirmation, we might try to cancel the same transaction multiple times (fix: keep an in memory array of pending cancellations). Wdyt @bajtos?

juliangruber: I've been running this script on my machine for a bit and the selected approach seems to work fine :)

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+
+const ROUND_LENGTH_MS = ms('20 minutes')
+
+/**
+  * @param {string} f4addr
+  * @returns 1000 oldest messages
+  */
+async function getMessagesInMempool (f4addr) {
+  const res = await pRetry(
+    () => fetch(`https://filfox.info/api/v1/message/mempool/filtered-list?address=${f4addr}&pageSize=1000`),
```
bajtos, line 15: Should we uri-encode the address, just to be safe?
```suggestion
    () => fetch(`https://filfox.info/api/v1/message/mempool/filtered-list?address=${encodeURIComponent(f4addr)}&pageSize=1000`),
```

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+    }[];
+  }}
+  */
+  const { messages } = /** @type {any} */(await res.json())
+  return messages
+}
+
+/**
+ * @returns {Promise<{
+  "cid": string;
```
bajtos, line 49: _Nitpick - feel free to ignore._

We usually don't put property names into (double quotes).

```suggestion
  cid: string;
```

The same comment applies to all properties below too.

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+      "baseFeeBurn": string;
+      "overEstimationBurn": string;
+      "minerPenalty": string;
+      "minerTip": string;
+      "refund": string;
+    },
+  }>}
+*/
+async function getRecentSendMessage () {
+  let res = await fetch('https://filfox.info/api/v1/message/list?method=Send')
```
bajtos, line 75: Should we wrap this HTTP request in `pRetry`?

In other words, why do we need to implement retry for the request `https://filfox.info/api/v1/message/mempool/filtered-list` but not for the request for `https://filfox.info/api/v1/message/list`?

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+  if (!res.ok) {
+    throw new Error(`Filfox request failed with ${res.status}: ${(await res.text()).trimEnd()}`)
+  }
+  const body = /** @type {any} */(await res.json())
+  assert(body.messages.length > 0, '/message/list returned an empty list')
+  const sendMsg = body.messages.find(m => m.method === 'Send')
+  assert(!!sendMsg, 'No Send message found in the recent committed messages')
+  const cid = sendMsg.cid
+
+  res = await fetch(`https://filfox.info/api/v1/message/${cid}`)
```
bajtos, line 85: Ditto - should we add `pRetry`?

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+  return /** @type {any} */(await res.json())
+}
+
+const cancelStuckTxs = async ({ walletDelegatedAddress, signer }) => {
+  console.log('Checking for stuck transactions...')
+
+  const messages = await getMessagesInMempool(walletDelegatedAddress)
+  const txsToCancel = messages.filter(m => {
+    return new Date(m.createTimestamp * 1000) <
+      new Date(Date.now() - (2 * ROUND_LENGTH_MS))
```
bajtos, line 99: A simpler alternative to consider:

```suggestion
    return m.createTimestamp * 1000 <
      (Date.now() - 2 * ROUND_LENGTH_MS)
```

lib/cancel-stuck-txs.js
bajtos: I am happy with the approach you implemented 👍🏻

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+      })
+      console.log(
+        `Waiting for receipt of replacement ${replacementTx.hash} for ${tx.cid}`
+      )
+      await replacementTx.wait()
+      console.log(`Replaced ${tx.cid}`)
+    } catch (err) {
+      console.error(err)
+      Sentry.captureException(err)
+    }
```
bajtos, line 146: This closure function has more than 20 lines of code. Would you mind extracting it into a standalone function?

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+    const oldGasPremium = Number(tx.gasPremium)
+    const nonce = tx.nonce
+
+    console.log(`Replacing ${tx.cid}...`)
+    try {
+      const replacementTx = await signer.sendTransaction({
+        to: signer.address,
+        value: 0,
+        nonce,
+        gasLimit: Math.ceil(gasUsed * 1.1),
```
bajtos, line 134: Should we extract the magic constant `1.1` into a named constant defined at the top of the file to make it more clear what knobs we can turn to change the behaviour of the auto-cancellation process?

Also, could you please add a code comment to explain to future readers how to choose this value?

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+        maxFeePerGas: gasFeeCap,
+        maxPriorityFeePerGas: Math.ceil(oldGasPremium * 1.252)
```
bajtos, line 136: Ditto. Should we extract the magic constant `1.252` into a named constant defined at the top of the file to make it more clear what knobs we can turn to change the behaviour of the auto-cancellation process?

I am also wondering how you arrived at the value `1.252`. Why not `1.2` or `1.3`? Could you please add a code comment to explain it to future readers?

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+      })
+      console.log(
+        `Waiting for receipt of replacement ${replacementTx.hash} for ${tx.cid}`
+      )
+      await replacementTx.wait()
+      console.log(`Replaced ${tx.cid}`)
```
bajtos, line 142: Could you please include both `tx.cid` and `replacementTx.hash` in this log message? If it takes >10 minutes for a replacement transaction to be confirmed, then the previous log message related to this operation will be long time lost among other message logged in the meantime.

lib/cancel-stuck-txs.js
```diff
@@ -0,0 +1,171 @@
...
+      didCancelTxs = await cancelStuckTxs({
+        walletDelegatedAddress,
+        signer
+      })
+    } catch (err) {
+      console.error(err)
+      Sentry.captureException(err)
+    }
+    if (!didCancelTxs) {
+      await timers.setTimeout(60_000)
```
bajtos, line 168: Should we use `ms('1 min')` for consistency? Perhaps extract this into a top-level constant.

bajtos (commented):
The implementation looks great at the high level 👏🏻 

I left a bunch of minor comments to consider.

The biggest issue I see is the lack of test coverage. I know this code is difficult to test because it interacts with live blockchain. What do you think about adding a smoke test calling `cancelStuckTxs()`? Or is that going to take too long to finish? 

Alternatively, we can at least add tests calling `getMessagesInMempool()` and `getRecentSendMessage()` to ensure our integration with the Filfox Explorer API works as expected. (The API can change in the future.)

We can also land this without tests to ship the improvement quickly, and then add the tests later this week. Let me know what you prefer!

juliangruber:
> What do you think about adding a smoke test calling cancelStuckTxs()? Or is that going to take too long to finish?

In my tests this sometimes took 13 minutes to complete.

> Alternatively, we can at least add tests calling getMessagesInMempool() and getRecentSendMessage() to ensure our integration with the Filfox Explorer API works as expected. (The API can change in the future.)

Good idea! I'll add those

lib/cancel-stuck-txs.js
juliangruber: I don't see why, the address comes from our ethers signer, which we're trusting everywhere else too

lib/cancel-stuck-txs.js
juliangruber: I have copy pasted this from your PR, will adjust

lib/cancel-stuck-txs.js
juliangruber: I added retrying to that other endpoint because I saw it fail, this one I hadn't seen fail. I'm +-0 on adding it, wdyt?

lib/cancel-stuck-txs.js
juliangruber: I didn't come up with this number, I don't know the reason why it's 1.1 tbh

lib/cancel-stuck-txs.js
juliangruber: Refactored using `ms()` and moved to the top level 👍

lib/cancel-stuck-txs.js
bajtos: I don't have a strong opinion. Adding `pRetry` should be an easy change we can do in the future when the need arises, right?

test/cancel-stuck-txs.test.js
```diff
@@ -0,0 +1,22 @@
...
+import {
+  getMessagesInMempool,
+  getRecentSendMessage
+} from '../lib/cancel-stuck-txs.js'
+
+describe('cancel stuck transactions', () => {
+  describe('getMessagesInMempool(addr)', () => {
+    it('should return messages in mempool', async () => {
+      const messages = await getMessagesInMempool(
+        '0x000000000000000000000000000000000000dEaD'
```
bajtos, line 11: Can we use a real address, please? For example, the current f4 address that is used by the spark-evaluate service. We can hard-code this address in this file, no need to create a wallet.

bajtos (changes requested):
Almost there!

lib/cancel-stuck-txs.js
juliangruber: Tbh, 1.1 doesn't make sense to me, wrt Riba's comment, which states:

> you also must LOWER your gas-limit to match the new gas requirements of the message you are sending:
if you do not do so, the network will penalize you for lying ( you said this will cost X gas but
used 1/10th of X to execute: no good )

Is increasing the gas limit value actually lowering the gas limit?

test/cancel-stuck-txs.test.js
juliangruber: We could, but it most of the time won't have any messages in the mempool. What difference does it make?

bajtos: I don't see any transactions for 0x000000000000000000000000000000000000dEaD, maybe this address is handled in a special way?

https://filfox.info/en/address/0x000000000000000000000000000000000000dEaD

How about using the spark-publish address? That one should have one transaction in the mempool quite often, or not?

juliangruber: Well, but what would it test? Are you proposing that if one message is returned we check the shape of it? I don't yet understand what would change if we changed the address to one that more likely has messages in the mempool

lib/cancel-stuck-txs.js
bajtos: Hmm, this is confusing, isn't it? 

Here is my understanding: We need to lower the gas limit from the limit calculated for the expensive smart contract call transaction to the limit that's reasonable for a cheap SEND transaction.

To do so, it should be enough to set `gasLimit=recentGasUsed` (set the limit to the real amount of gas used by a recent SEND transaction).

Then there are two questions:
1. Why do we need to increase this value?
2. Why 10% and not a different ratio?

I _think_ that the answer is to account for a possible rise in the base fee since the last SEND transaction and the 10% value is a heuristic - a tradeoff between having a high confidence (but paying too much) vs paying just enough with the risk of the replacement transaction getting stuck again.

WDYT?

test/cancel-stuck-txs.test.js
bajtos: Yeah, it would be ideal to test the shape of the returned message. OTOH, if the response is empty most of the time, then I agree this does not matter that much.

bajtos (approved):
👏🏻 

I think this is good enough to get landed. Please remember to fix the failing TypeScript check - you forgot to apply the rename in the place calling `cancelTx()`.

Screenshot 2024-08-27 at 15 33 42

We should eventually get a better understanding of the `1.1` constant (see https://github.com/filecoin-station/spark-evaluate/pull/329#discussion_r1732743601), but that's not a blocker for landing this PR.

test/cancel-stuck-txs.test.js
juliangruber: I think we can only have a message there reliably if we switch to testnet and create a new tx as part of the test

juliangruber:
> We should eventually get a better understanding of the 1.1 constant (see https://github.com/filecoin-station/spark-evaluate/pull/329#discussion_r1732743601), but that's not a blocker for landing this PR.

I very much agree :|

lib/cancel-stuck-txs.js
juliangruber: Hmm yeah that could be, it makes sense like this I think

bajtos:
> ## Suspect Issues
> 
> * ‼️ **Error: Filfox request failed with 502: ** `getMessagesInMempool(cancel-stuck-txs)` [View Issue](https://space-meridian.sentry.io/issues/5766240201/?referrer=github-pr-bot)
> 

@juliangruber Could you PTAL?

juliangruber:
On it

juliangruber:
https://github.com/filecoin-station/spark-evaluate/pull/332

PR #333, opened 2024-08-28
Remove unresponsive chainup endpoint

juliangruber:
Unfortunately this endpoint has been unresponsive for quite some time now -> we're back to glif-only endpoints


bajtos (approved):
:shipit: 

> Unfortunately this endpoint has been unresponsive for quite some time now -> we're back to glif-only endpoints

😢

PR #336, opened 2024-08-30
feat!: per-miner & per-client daily deal stats

bajtos:
Rework the `daily_deals` table into the following schema:

For each `(day, miner_id, client_id)`, we want to know the following numbers (counts):
 - `tested`: (NEW) total deals tested
 - `index_majority_found`: (NEW) deals where we found a majority agreeing on the same result of the IPNI query
 - `indexed`: deals announcing retrievals to IPNI (HTTP or Graphsync retrievals)
 - `indexed_http`: (NEW) deals announcing HTTP retrievals to IPNI
 - `retrieval_majority_found`: (NEW) deals where we found a majority agreeing on the same result of the retrieval request
 - `retrievable`: deals where the majority agrees the content can be retrieved

BREAKING CHANGE: spark-stats endpoints consuming this table need to change queries to use `tested` instead of `total`.

Links:
- https://www.notion.so/spacemeridian/Move-Spark-score-from-individual-retrievals-to-retrievable-deals-09dc84797aea48ee84dd7a78f09a8b49?d=356ac20afa604231be63e2f30364dd3a#e81f9849d9a34006a0013830e409843d

Remaining work:
- [x] add tests for the new columns
- [x] add index to speed up date-based lookups (`INDEX ON daily_deals (day)`)


lib/typings.d.ts
```diff
@@ -11,6 +11,7 @@ export {
 export interface RetrievalTask {
   cid: string;
   minerId: string;
+  clients?: string[];
```
bajtos, line 14: See https://github.com/filecoin-station/spark-api/pull/387

juliangruber:
> BREAKING CHANGE: spark-stats endpoints consuming this table need to change queries to use tested instead of total.

Can we use this as an opportunity to fix this coupling, by adding an HTTP endpoint to `spark-evaluate` and letting `spark-stats` consume that, instead of querying its db?

lib/public-stats.js
```diff
@@ -101,37 +102,108 @@ const updateIndexerQueryStats = async (pgClient, committees) => {
...
+   * indexed:  number;
+   * indexed_http: number;
+   * retrievable: number;
+   * }>>} */
+  const minerClientDealStats = new Map()
   for (const c of committees) {
-    total++
+    const { minerId, cid } = c.retrievalTask
+    let clients = findDealClients(minerId, cid)
+    if (!(clients && clients.length)) {
```
juliangruber, line 119:
```suggestion
    if (!clients || !clients.length) {
```

I find this easier to reason about, but don't mind keeping the original version if you prefer it

lib/public-stats.js
```diff
@@ -101,37 +102,108 @@ const updateIndexerQueryStats = async (pgClient, committees) => {
...
+      console.warn(`Invalid retrieval task (${minerId}, ${cid}): no deal clients found. Using a placeholder value 'no-clients'.`)
```
juliangruber, line 120: what if we instead kept the `clients` array empty?

juliangruber (changes requested):
Please see https://github.com/filecoin-station/spark-evaluate/pull/336#issuecomment-2322959419

bajtos:
> > BREAKING CHANGE: spark-stats endpoints consuming this table need to change queries to use tested instead of total.
> 
> Can we use this as an opportunity to fix this coupling, by adding an HTTP endpoint to `spark-evaluate` and letting `spark-stats` consume that, instead of querying its db?

spark-evaluate does not expose any HTTP interface right now. Changing that is way beyond the scope of this work, IMO.

IIRC, the direction we wanted to take is to improve spark-evaluate to publish the evaluation results to IPFS & commit the CID to the smart contract and then move all code building publicly-retrievable stats from spark-evaluate to spark-stats repository.

lib/public-stats.js
bajtos: Then, we will not insert any records into the table because we need both the miner & client for that. As a result, we will exclude deals where we don't have any client info from the DRS calculation.

Having written that, it should never happen that there is a task with no deal clients. 

The table `retrievable_deals` which tracks deals eligible for retrievability testing uses `(cid, miner_id)` as the primary key:

https://github.com/filecoin-station/spark-api/blob/2712b4e04c37f6759fb832701b3b6b2afb9fa001/migrations/052.do.retrievable-deals-clients.sql#L5

Per-round retrieval tasks are defined as a random sample from that table, so each task must have at least one deal client:

https://github.com/filecoin-station/spark-api/blob/2712b4e04c37f6759fb832701b3b6b2afb9fa001/api/lib/round-tracker.js#L260-L273

I want to handle this unlikely edge case to avoid losing measurements. Even if we cannot link them to any deal client, we can still link them to a miner (an SP).

juliangruber: > Having written that, it should never happen that there is a task with no deal clients.

Shall we log this situation to Sentry then?

juliangruber:
What about these:

- `index_majority_found`
- `retrieval_majority_found`

I think these make sense because they stand for "majority found in the index process" and "majority found in the retrieval process". Strictly speaking it's "index resolution", so should be `index_resolution_majority_found`, but I don't think this level of precision is necessary

lib/public-stats.js
bajtos: I was half expecting this comment 😄  I am fine either way.

juliangruber: This is also happening to me!

lib/public-stats.js
bajtos: Added Sentry logging in 6424760

bajtos:
The PR is ready for final review & landing. @juliangruber PTAL 🙏🏻

lib/public-stats.js
```diff
@@ -104,37 +106,134 @@ const updateIndexerQueryStats = async (pgClient, committees) => {
...
+        }
+        clientDealStats.set(clientId, stats)
+      }
+
+      stats.tested++
+
+      const evaluation = c.evaluation
+      if (!evaluation) continue
+
+      if (evaluation.indexerResult !== 'COMMITTEE_TOO_SMALL' && evaluation.indexerResult !== 'MAJORITY_NOT_FOUND') {
```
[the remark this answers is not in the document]
bajtos: I see your point; I was also not entirely comfortable when I was writing this.

I am afraid the list of positive statuses is long and open-ended; see https://github.com/filecoin-station/spark-evaluate/blob/e4323a8938a781334b0107399fa6ae02a5afb4d3/lib/typings.d.ts#L48-L67

I guess this tells us we need more information for each evaluation field - not only the result but also a flag telling us whether a majority was found.

lib/public-stats.js
bajtos: What kind of inconsistency do you have in mind?

If we abort the operation, we discard measurements that provide valid information about a storage provider.

Having written that, we should never get an empty list of clients, so I am fine with skipping such a committee. We will get a Sentry report in the unlikely case that this happens, I think that shoud be good enough 👍🏻

lib/public-stats.js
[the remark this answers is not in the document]
juliangruber: Or we split up the result: 1. Was a majority found? 2. What is the outcome?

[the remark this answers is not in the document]
bajtos: Yeah, that would be easy to do with enum types in Rust:

```rs
enum RetrievalResult {
  IpniNotQueried,
  HttpError(uint16),
  Timeout,
  // etc.
}

enum NoMajorityCause {
  CommiteeTooSmall,
  NoMajorityFound
}

enum CommitteeRetrievalResult {
  MajorityFound(RetrievalResult),
  NoMajority(NoMajorityCause)
}
```

I don't know how to define such type in TS/JS, do you know?

lib/public-stats.js
bajtos: See c3f43f6

lib/public-stats.js
```diff
@@ -104,37 +106,134 @@ const updateIndexerQueryStats = async (pgClient, committees) => {
...
+
+      if (evaluation.indexerResult === 'OK' || evaluation.indexerResult === 'HTTP_NOT_ADVERTISED') {
+        stats.indexed++
+      }
+
+      if (evaluation.indexerResult === 'OK') {
+        stats.indexed_http++
+      }
+
+      if (evaluation.retrievalResult !== 'COMMITTEE_TOO_SMALL' && evaluation.indexerResult !== 'MAJORITY_NOT_FOUND') {
```
bajtos, line 172: BUG! The second condition is incorrectly testing `indexerResult` instead of `retrievalResult`.

lib/public-stats.js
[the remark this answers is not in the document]
bajtos: Addressed in d67526348 - I added `evaluation.hasIndexMajority` and `evaluation.hasRetrievalMajority` fields.

test/public-stats.test.js
```diff
@@ -467,7 +468,7 @@ describe('public-stats', () => {
           'SELECT day::TEXT, tested, retrieval_majority_found, retrievable FROM daily_deals'
         )
         assert.deepStrictEqual(created, [
-          { day: today, tested: 4, retrieval_majority_found: 3, retrievable: 1 }
+          { day: today, tested: 4, retrieval_majority_found: 2, retrievable: 1 }
```
bajtos, line 471: This change was needed after I fixed the bug described in https://github.com/filecoin-station/spark-evaluate/pull/336#discussion_r1749920681

test/public-stats.test.js
```diff
@@ -501,7 +503,7 @@ describe('public-stats', () => {
           'SELECT day::TEXT, tested, retrieval_majority_found, retrievable FROM daily_deals'
         )
         assert.deepStrictEqual(created, [
-          { day: today, tested: 4 + 3, retrieval_majority_found: 3 + 2, retrievable: 1 + 1 }
+          { day: today, tested: 4 + 3, retrieval_majority_found: 2 + 2, retrievable: 1 + 1 }
```
bajtos, line 506: This change was needed after I fixed the bug described in https://github.com/filecoin-station/spark-evaluate/pull/336#discussion_r1749920681

lib/public-stats.js
bajtos: I fixed this in [d675263](https://github.com/filecoin-station/spark-evaluate/pull/336/commits/d67526348fd5689d734b649fafa2260f1a26bfff)

bajtos:
@juliangruber LGTY now?

lib/public-stats.js
juliangruber: If `'no-clients'` makes it into the DB, I would consider that an inconsistency: You expect a client to be there if a row / column value is there.

juliangruber (approved):
Great work, Miro!!

PR #337, opened 2024-08-30
Add use `FallbackProvider` to improve reliability. Closes #281

### LINKED ISSUE #281: Use FallbackProvider to improve reliability

juliangruber:
Closes #281


lib/ie-contract.js
```diff
@@ -1,19 +1,16 @@
...
 export const createMeridianContract = async (contractAddress = SparkImpactEvaluator.ADDRESS) => {
-  const fetchRequest = new ethers.FetchRequest(RPC_URL)
-  fetchRequest.setHeader('Authorization', rpcHeaders.Authorization || '')
-  const provider = new ethers.JsonRpcProvider(
-    fetchRequest,
-    null,
-    {
+  const provider = new ethers.FallbackProvider(rpcUrls.map(url => {
+    const fetchRequest = new ethers.FetchRequest(url)
+    fetchRequest.setHeader('Authorization', `Bearer ${GLIF_TOKEN}`)
```
bajtos, line 8: Is it okay to send our GLIF_TOKEN to other RPC API providers?

bajtos (approved):

lib/ie-contract.js
juliangruber: We also do this in spark-api: https://github.com/filecoin-station/spark-api/blob/main/api/lib/ie-contract.js#L7

We could make this conditional (`if (url.includes('glif'))`), although I also don't see an immediate risk with this.

bajtos: Fair enough 👍🏻 

:shipit:

PR #338, opened 2024-08-30
Add integrate `spark-rewards`

juliangruber:
https://github.com/space-meridian/roadmap/issues/161
Closes https://github.com/filecoin-station/spark-evaluate/issues/364


bin/dry-run.js
```diff
@@ -121,25 +122,29 @@ for (const cid of measurementCids) {
...
 
 const started = Date.now()
 const { ignoredErrors } = await evaluate({
   roundIndex,
   round,
   fetchRoundDetails,
-  ieContractWithSigner,
+  ieContract,
+  signer,
+  fetch: mockedFetch,
```
bajtos, line 147: IIRC, evaluate calls fetch at other places, e.g. to fetch round details here:

https://github.com/filecoin-station/spark-evaluate/blob/3f93c1289bafaa5a970397e9f9b86e0b6a4176ff/lib/spark-api.js#L18

Let's design & mock this dependency at a higher level.

I propose the following interface:

```ts
function submitEvaluationScores(
  participants: string[],
  values: bigint[]
) => Promise<void>
```

Alternatively:

```ts
function submitEvaluationScores(
  participantScores: Record<string, bigint>
) => Promise<void>
```

(BTW, why are we using an Object and not a Map for the participant/score mapping?)

This way, we can have two implementation:
 - a no-op implementation for the dry run, this means the dry run script does not need any burner wallet
 - a production implementation that uses the `signer` initialised from the seed phrase

In the future, it will be easier to swap the spark-rewards implementation for a different one using IPC or other L2 network, as the code inside `evaluate()` will stay the same.

bajtos (changes requested):
See the comment below.

GitHub also shows a bunch of TypeScript errors. Did you run the `npm test`?

juliangruber:
> See the comment below.
> 
> GitHub also shows a bunch of TypeScript errors. Did you run the `npm test`?

This is a draft and I hadn't requested a review yet :) Also see the commit title

bin/dry-run.js
juliangruber: Yeah, I was also thinking about introducing a new function for this. Mocking `fetch` isn't fun. +1

> (BTW, why are we using an Object and not a Map for the participant/score mapping?)

I don't think there's a particular reason for this, and can be refactored.

test/evaluate.js
```diff
@@ -55,42 +55,33 @@ describe('evaluate', async function () {
...
-    const setScoresCalls = []
-    const ieContractWithSigner = {
-      async setScores (roundIndex, participantAddresses, scores) {
-        setScoresCalls.push({ roundIndex, participantAddresses, scores })
-        return { hash: '0x234' }
-      },
+    const submitScoresCalls = []
+    const submitScores = async (participantAddresses, scores) => {
+      submitScoresCalls.push({ participantAddresses, scores })
+    }
```
bajtos, line 61: Can you preserve the name `setScoresCall` to reduce the code churn?

I am surprised we don't need to include the `roundIndex` when submitting scores. Could the lack of `roundIndex` make debugging more difficult for us?

How do you feel about keeping the `setScores` signature in the interfaces, i.e. modify `submitScores` to require `(roundIndex, participantAddresses, scores)`, even if the actual implementation calling spark-rewards does not use the round index? 

Since it's something used by the previous version, I think we may need to have it again in the future. I know it's just a speculation violating YAGNI principle, so 🤷🏻  What I like most is that if we preserve the signature then there will be much fewer changes in the existing tests.

bajtos (changes requested):
Great work!

I have a few comments to discuss, PTAL below.

bajtos:
@juliangruber could you please remind me what's the plan for paying out rewards tracked by spark-rewards and how are we going to handle the migration?

bin/spark-evaluate.js
```diff
@@ -30,36 +29,45 @@ await migrateWithPgConfig({ connectionString: DATABASE_URL })
...
   return pgClient
 }
 
-const pgClient = await createPgClient()
-const stuckTransactionsCanceller = createStuckTransactionsCanceller({
-  pgClient,
-  // Bypass NonceManager as we need to cancel transactions with the same nonce
-  signer: wallet
-})
+const submitScores = async (participants, values) => {
```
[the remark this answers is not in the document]
juliangruber: I've refactored the function, I think adding tests is lower priority due to the simplicity of the function, therefore I added a comment about adding tests in the future instead

test/evaluate.js
juliangruber: > Can you preserve the name setScoresCall to reduce the code churn?

This would mean we have a function `submitScores` that pushes into `setScoresCalls`. That doesn't make sense. It's only coherent if we also rename `submitScores` to `setScores` everywhere. I'm ok with that, wdyt?

> I am surprised we don't need to include the roundIndex when submitting scores. Could the lack of roundIndex make debugging more difficult for us?

`spark-rewards` has no notion of rounds. What would we need the round index for?

> How do you feel about keeping the setScores signature in the interfaces, i.e. modify submitScores to require (roundIndex, participantAddresses, scores), even if the actual implementation calling spark-rewards does not use the round index?

We don't need that right now so I'm against it and think we shouldn't spend time discussing it. If you feel strongly I'll just change it.

juliangruber:
> @juliangruber could you please remind me what's the plan for paying out rewards tracked by spark-rewards

Docs for reference: https://spacemeridian.notion.site/Meridian-Off-chain-scheduled-rewards-f9480ef009464ecfaf02a4f752cc4d70#6b1b57295e3046a395085473027d8942

The steps are:
1. `GET /scores` on `spark-rewards`
2. `addBalances()` on the IE
3. `POST /paid` on `spark-rewards`

We should write a CLI for this, as doing this manually is brittle. I'll do that next.

> and how are we going to handle the migration - what are the next steps after landing this PR up to the first payout of rewards?

We don't need to migrate anything, as calling `releaseRewards()` on the IE will release rewards both from before and after merging this PR. I will take care of updating the docs.

bin/spark-evaluate.js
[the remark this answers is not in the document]
bajtos: > I've refactored the function, 

Thank you!

> I think adding tests is lower priority due to the simplicity of the function, therefore I added a comment about adding tests in the future instead

I am curious - have you ever run this code, or will we execute it for the first time in production? 

We did the latter recently (https://github.com/filecoin-station/spark-evaluate/pull/363, landed on Sept 24th) and there was a bug we did not discover & fix until the next day (https://github.com/filecoin-station/spark-evaluate/pull/368).

Please watch the production after landing this PR until the first round is evaluated & scores are submitted to ensure the code works as intended.

test/evaluate.js
bajtos: > > Can you preserve the name setScoresCall to reduce the code churn?
> 
> This would mean we have a function `submitScores` that pushes into `setScoresCalls`. That doesn't make sense. It's only coherent if we also rename `submitScores` to `setScores` everywhere. I'm ok with that, wdyt?

_rename `submitScores` to `setScores` everywhere_ sounds good to me 👍🏻 

> > I am surprised we don't need to include the roundIndex when submitting scores. Could the lack of roundIndex make debugging more difficult for us?
> 
> `spark-rewards` has no notion of rounds. What would we need the round index for?

For troubleshooting - e.g. to detect a duplicate submission of scores for the same round, or a round with no scores submitted.

It also allows us to implement idempotency in spark-rewards: we can add pRetry to spark-evaluate and drop duplicate scores for the same round on the spark-rewards side.

> > How do you feel about keeping the setScores signature in the interfaces, i.e. modify submitScores to require (roundIndex, participantAddresses, scores), even if the actual implementation calling spark-rewards does not use the round index?
> 
> We don't need that right now so I'm against it and think we shouldn't spend time discussing it. If you feel strongly I'll just change it.

Ack!

bajtos:
> > @juliangruber could you please remind me what's the plan for paying out rewards tracked by spark-rewards
> 
> Docs for reference: https://spacemeridian.notion.site/Meridian-Off-chain-scheduled-rewards-f9480ef009464ecfaf02a4f752cc4d70#6b1b57295e3046a395085473027d8942
> 
> The steps are:
> 
> 1. `GET /scores` on `spark-rewards`
> 2. `addBalances()` on the IE
> 3. `POST /paid` on `spark-rewards`
> 
> We should write a CLI for this, as doing this manually is brittle. I'll do that next.
> 
> > and how are we going to handle the migration - what are the next steps after landing this PR up to the first payout of rewards?
> 
> We don't need to migrate anything, as calling `releaseRewards()` on the IE will release rewards both from before and after merging this PR. I will take care of updating the docs.

Perfect!

Do we need to make any changes inside Station Core and/or Station Desktop to let the chart in the Station Desktop UI to show the correct amount of scheduled rewards? This amount can be a combination of scheduled rewards tracked by the smart contract and by spark-rewards.

bajtos (approved):
The new version is good enough 👍🏻 

Please take a look at https://github.com/filecoin-station/spark-evaluate/pull/338#discussion_r1789938336 before landing.

**EDIT**

Also please consider the impact on Station Desktop UI before landing, see https://github.com/filecoin-station/spark-evaluate/pull/338#issuecomment-2396529371.

juliangruber:
Converted to draft as these projects should have `spark-rewards` integrated first:
- [ ] `core`
- [ ] `spark-stats`
- [ ] `telegraf`

juliangruber:
Blocked by https://github.com/filecoin-station/spark-rewards/issues/13

test/evaluate.js
juliangruber: > rename submitScores to setScores everywhere sounds good to me 👍🏻

Done!

> For troubleshooting - e.g. to detect a duplicate submission of scores for the same round, or a round with no scores submitted.
>
> It also allows us to implement idempotency in spark-rewards: we can add pRetry to spark-evaluate and drop duplicate scores for the same round on the spark-rewards side.

I opened a new issue for this, I don't see it as release blocking: https://github.com/filecoin-station/spark-rewards/issues/15

PR #341, opened 2024-08-30
feat: rework top_measurement_participants_yesterday_mv  to use the new schema

PatrickNercessian:
Relates to #311 and https://github.com/space-meridian/roadmap/issues/132

### DIFF
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -312,7 +312,6 @@ export const refreshDatabase = async (
   {
     functionsToRun = [
       updateTopMeasurementParticipants,
-      updateTopMeasurementParticipantsView,
       aggregateAndCleanUpRecentData,
       updateMonthlyActiveStationCount
     ]
@@ -336,39 +335,17 @@ export const refreshDatabase = async (
  * @param {import('pg').Client} pgClient
  */
 export const updateTopMeasurementParticipants = async (pgClient) => {
-  const mvTableName = 'top_measurement_participants_yesterday_mv'
   try {
     await pRetry(async () => {
-      await pgClient.query(`REFRESH MATERIALIZED VIEW ${mvTableName}`)
+      await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
     }, {
       retries: 3,
       onFailedAttempt: (error) => {
-        console.warn(`Attempt to refresh ${mvTableName} failed: ${error.message}. Retrying...`)
+        console.warn(`Attempt to refresh materialized view failed: ${error.message}. Retrying...`)
       }
     })
   } catch (err) {
-    console.error(`Error refreshing ${mvTableName}`, err)
-    Sentry.captureException(err)
-  }
-}
-
-/**
- * @param {import('pg').Client} pgClient
- */
-export const updateTopMeasurementParticipantsView = async (pgClient) => {
-  const mvTableName = 'top_measurement_participants_yesterday_view'
-  try {
-    await pRetry(async () => {
-      await pgClient.query(`REFRESH MATERIALIZED VIEW ${mvTableName}`)
-      console.log(`Refreshed ${mvTableName}`)
-    }, {
-      retries: 3,
-      onFailedAttempt: (error) => {
-        console.warn(`Attempt to refresh ${mvTableName} failed: ${error.message}. Retrying...`)
-      }
-    })
-  } catch (err) {
-    console.error(`Error refreshing ${mvTableName}`, err)
+    console.error('Error refreshing top measurement participants', err)
     Sentry.captureException(err)
   }
 }
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -312,6 +312,7 @@ export const refreshDatabase = async (
   {
     functionsToRun = [
       updateTopMeasurementParticipants,
+      updateTopMeasurementParticipantsView,
       aggregateAndCleanUpRecentData,
       updateMonthlyActiveStationCount
     ]
@@ -335,17 +336,39 @@ export const refreshDatabase = async (
  * @param {import('pg').Client} pgClient
  */
 export const updateTopMeasurementParticipants = async (pgClient) => {
+  const mvTableName = 'top_measurement_participants_yesterday_mv'
   try {
     await pRetry(async () => {
-      await pgClient.query('REFRESH MATERIALIZED VIEW top_measurement_participants_yesterday_mv')
+      await pgClient.query(`REFRESH MATERIALIZED VIEW ${mvTableName}`)
     }, {
       retries: 3,
       onFailedAttempt: (error) => {
-        console.warn(`Attempt to refresh materialized view failed: ${error.message}. Retrying...`)
+        console.warn(`Attempt to refresh ${mvTableName} failed: ${error.message}. Retrying...`)
       }
     })
   } catch (err) {
-    console.error('Error refreshing top measurement participants', err)
+    console.error(`Error refreshing ${mvTableName}`, err)
+    Sentry.captureException(err)
+  }
+}
+
+/**
+ * @param {import('pg').Client} pgClient
+ */
+export const updateTopMeasurementParticipantsView = async (pgClient) => {
+  const mvTableName = 'top_measurement_participants_yesterday_view'
+  try {
+    await pRetry(async () => {
+      await pgClient.query(`REFRESH MATERIALIZED VIEW ${mvTableName}`)
+      console.log(`Refreshed ${mvTableName}`)
+    }, {
+      retries: 3,
+      onFailedAttempt: (error) => {
+        console.warn(`Attempt to refresh ${mvTableName} failed: ${error.message}. Retrying...`)
+      }
+    })
+  } catch (err) {
+    console.error(`Error refreshing ${mvTableName}`, err)
     Sentry.captureException(err)
   }
 }
--- a/migrations/014.do.top-measurement-participants-mv.sql
+++ b/migrations/014.do.top-measurement-participants-mv.sql
@@ -1,0 +1,36 @@
+CREATE MATERIALIZED VIEW top_measurement_participants_yesterday_view AS
+WITH yesterday_station_and_measurement_counts AS (
+  SELECT
+    day,
+    participant_id,
+    COUNT(DISTINCT station_id) AS station_count,
+    SUM(accepted_measurement_count) AS accepted_measurement_count
+  FROM
+    recent_station_details
+  WHERE
+    day = CURRENT_DATE - INTERVAL '1 day'
+  GROUP BY
+    day, participant_id
+),
+yesterday_subnet_counts AS (
+  SELECT 
+    day,
+    participant_id,
+    COUNT(DISTINCT subnet) AS subnet_count
+  FROM
+    recent_participant_subnets
+  WHERE
+    day = CURRENT_DATE - INTERVAL '1 day'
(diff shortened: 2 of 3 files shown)


lib/platform-stats.js
```diff
@@ -335,17 +336,39 @@ export const refreshDatabase = async (
...
-    console.error('Error refreshing top measurement participants', err)
+    console.error(`Error refreshing ${mvTableName}`, err)
+    Sentry.captureException(err)
+  }
+}
+
+/**
+ * @param {import('pg').Client} pgClient
+ */
+export const updateTopMeasurementParticipantsView = async (pgClient) => {
```
PatrickNercessian, line 358: If we were going to keep both the old schema and the new one, it would make sense to combine these two methods (since they're basically the exact same code)

However, we're planning to delete the old schema and its code soon, so I thought it might not be worth combining them (and then handling SQL query safeness with the inserted table name)

What do you guys think?

migrations/014.do.top-measurement-participants-mv.sql
```diff
@@ -0,0 +1,36 @@
+CREATE MATERIALIZED VIEW top_measurement_participants_yesterday_view AS
```
PatrickNercessian, line 1: Was hard to pick a good name, since this exact thing already exists (in terms of the data stored, not how it retrieves it). Maybe you have a better approach in mind?

test/platform-stats.test.js
```diff
@@ -275,7 +276,7 @@ describe('platform-stats', () => {
         { ...VALID_MEASUREMENT, stationId: 'station1', participantAddress: '0x10', inet_group: 'subnet1', fraudAssessment: 'TASK_NOT_IN_ROUND' }
       ]
 
-      await updateStationsAndParticipants(pgClient, allMeasurements, participantsMap, today)
+      await updateStationsAndParticipants(pgClient, allMeasurements, participantsMap, { day: today })
```
PatrickNercessian, line 279: In creating the new test, I realized this was erroneously formatted before

lib/platform-stats.js
```diff
@@ -335,17 +336,39 @@ export const refreshDatabase = async (
  * @param {import('pg').Client} pgClient
  */
 export const updateTopMeasurementParticipants = async (pgClient) => {
+  const mvTableName = 'top_measurement_participants_yesterday_mv'
```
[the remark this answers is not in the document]
juliangruber: Got it, materialized view. I would like us to use as few not widely known abbreviations as possible

lib/platform-stats.js
PatrickNercessian: I had it that way before, but then we just have to directly write the table name in a string 3 separate times. And the string is internal to the function, not a real parameter, so no risk that I see unless a future change turns it into a parameter, which would be possible anyway.

But we can remove it if preferred, no hard preference from my end

lib/platform-stats.js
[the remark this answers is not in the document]
PatrickNercessian: materialized view (this one is already the existing table name)

migrations/014.do.top-measurement-participants-mv.sql
PatrickNercessian: The difference in retrieval is basically that we get the data from two tables now, `recent_station_details` and `recent_participant_subnets`, instead of just one before (`daily_stations`)

I can't think of a good name related to that though. Especially since this will last long-term, while the other will not, so any specifics about the differences will be lost to time (in one year, "Why did we name this "top_measurement_participants_yesterday_multiple_tables?" haha)

I like the `_deprecated` idea for the old one, but it would require changing the name for the existing table in this migration script though, and then everywhere else that uses it (including creating a new PR to change `spark-stats`)

bajtos: How about doing the simplest thing and appending `2` or `_v2` to the table name? It's not the prettiest thing to do, but it gives us a clear path forward if we ever need v3 or v4 of this.

PatrickNercessian: This sounds good to me. @juliangruber ?

bajtos, line 1: Do we actually need to keep both materialised views around?

I believe we are already populating the new tables with data. By the time we land this PR next week, we should have all the data required to create the "top measurement participants" view using the new approach.

What happens if this PR simply changes the definition of the existing view?
- No changes in spark-stats are needed; it will return data from the new view.
- The view will contain stats starting this week because we haven't backfilled the underlying tables yet.
- When the backfilling is done, the next refresh will backfill the data in the view too.

I think we can afford to not show historical data for a while, if it means less work & faster delivery of this feature.

Is there anything I missed?

migrations/014.do.top-measurement-participants-mv.sql
```diff
@@ -0,0 +1,36 @@
...
+SELECT 
+  ysmc.day,
+  p.participant_address,
+  ysmc.station_count,
+  ysmc.accepted_measurement_count,
+  ysc.subnet_count
+FROM yesterday_station_and_measurement_counts AS ysmc 
+JOIN yesterday_subnet_counts AS ysc ON ysmc.day = ysc.day AND ysmc.participant_id = ysc.participant_id
+JOIN participants p ON ysmc.participant_id = p.id
+ORDER BY ysmc.accepted_measurement_count DESC;
```
bajtos, line 36: Missing EOL at EOF

migrations/014.do.top-measurement-participants-mv.sql
PatrickNercessian: Ah great point, a bit embarrassed to have missed it haha. And then we avoid naming issues mentioned above too

bajtos: > And then we avoid naming issues mentioned above too

Yes, exactly!

BTW I run your query for today (removed `- INTERVAL '1 day'`) and got the following data (~7k participants):

https://gist.github.com/bajtos/611dcf3ed5e688a92dc386e7b5f0b6f5

bajtos: Oh, this materialised view shows data for yesterday only, right? So we don't need to worry about missing historical data because that data is not shown in this view anyway.

bajtos: See https://github.com/filecoin-station/spark-evaluate/pull/341#discussion_r1738941823 - I think we don't need the second materialised view and can simply redefine the existing one.

PatrickNercessian: Are you able to verify if it's the same data as the existing view?

test/platform-stats.test.js
```diff
@@ -196,39 +196,6 @@ describe('platform-stats', () => {
       const { rows } = await pgClient.query('SELECT station_id, day::TEXT FROM daily_stations')
       assert.deepStrictEqual(rows, [{ station_id: VALID_STATION_ID, day: today }])
     })
-
-    it('updates top measurements participants yesterday materialized view', async () => {
```
PatrickNercessian, line 200: Basically just moved this test to a different describe block, since it no longer has to do with daily_stations.

Plus some other minor changes to work with new schema

juliangruber (approved):

bajtos (changes requested):
I am afraid there are subtle differences between the output of the old and the new query.

**CURRENT**

```
            participant_address             | inet_group_count | station_count | accepted_measurement_count 
--------------------------------------------+------------------+---------------+----------------------------
 0xF7D8c74507863121dAF5D0d8b7501901718EfCd9 |              133 |           199 |                     120435
 0xCE53e6dEfbD1f0810BFcd691660ad311d0DB24f7 |              115 |           130 |                     102068
 0x7581D288f51622901eaB1d340be773241c53D025 |               99 |           100 |                      86489
(...)
```

**PROPOSED VERSION**

```
    day     |            participant_address             | station_count | accepted_measurement_count | subnet_count 
------------+--------------------------------------------+---------------+----------------------------+--------------
 2024-09-01 | 0xF7D8c74507863121dAF5D0d8b7501901718EfCd9 |           199 |                     120435 |          133
 2024-09-01 | 0xCE53e6dEfbD1f0810BFcd691660ad311d0DB24f7 |           130 |                     102068 |          115
 2024-09-01 | 0x7581D288f51622901eaB1d340be773241c53D025 |           100 |                      86489 |           99
```

Differences:
- The new view includes the `day` column. I think this is not necessarily a bad thing.
- `inet_group_count` is renamed to `subnet_count` - I think this will break consumers of this data

When I cleanse the data, the data returned by the current version and the new version seem to be the same except that participants with the same number of accepted measurements are ordered differently.

Full data:

- [diff](https://gist.githubusercontent.com/bajtos/c4b9313592cbcde6301e353f591a9a52/raw/ffdc37b97a30e803acc70aed67ded70583b86526/delta.diff)
- [current](https://gist.githubusercontent.com/bajtos/c4b9313592cbcde6301e353f591a9a52/raw/ffdc37b97a30e803acc70aed67ded70583b86526/old.txt)
- [proposed](https://gist.githubusercontent.com/bajtos/c4b9313592cbcde6301e353f591a9a52/raw/ffdc37b97a30e803acc70aed67ded70583b86526/new.txt)

migrations/014.do.top-measurement-participants-mv.sql
bajtos: See https://github.com/filecoin-station/spark-evaluate/pull/341#pullrequestreview-2275461145

PatrickNercessian:
@bajtos 

> I am afraid there are subtle differences between the output of the old and the new query.
> 
> * The new view includes the `day` column. I think this is not necessarily a bad thing.

Ah yes, forgot to mention this change. I did it intentionally, I think it's better to have this in the data, rather than having to guess which day the data corresponds to.

> * `inet_group_count` is renamed to `subnet_count` - I think this will break consumers of this data

Good catch, addressed in https://github.com/filecoin-station/spark-evaluate/pull/341/commits/2f7442aa679cc9e72e31d9aad31f3075d7a49d5f

> When I cleanse the data, the data returned by the current version and the new version seem to be the same except that participants with the same number of accepted measurements are ordered differently.

I think this makes sense to be happening. Theoretically it might even be different from subsequent runs of the same query, since we're ordering based on just the accepted_measurement_count. Do you think we should change that to make it more deterministic? We could order secondarily by `participant_address`, for example.

bajtos:
> @bajtos
> 
> > I am afraid there are subtle differences between the output of the old and the new query.
> > 
> > * The new view includes the `day` column. I think this is not necessarily a bad thing.
> 
> Ah yes, forgot to mention this change. I did it intentionally, I think it's better to have this in the data, rather than having to guess which day the data corresponds to.

👍🏻 

> > * `inet_group_count` is renamed to `subnet_count` - I think this will break consumers of this data
> 
> Good catch, addressed in [2f7442a](https://github.com/filecoin-station/spark-evaluate/commit/2f7442aa679cc9e72e31d9aad31f3075d7a49d5f)

Thanks!

> > When I cleanse the data, the data returned by the current version and the new version seem to be the same except that participants with the same number of accepted measurements are ordered differently.
> 
> I think this makes sense to be happening. Theoretically it might even be different from subsequent runs of the same query, since we're ordering based on just the accepted_measurement_count. Do you think we should change that to make it more deterministic? We could order secondarily by `participant_address`, for example.

Yes, I think it would be better to order secondarily by `participant_address`, but I also think this does not matter too much.

bajtos (approved):
👏🏻

PR #346, opened 2024-09-04
Fix priority fee below max fee

juliangruber:
https://github.com/ethers-io/ethers.js/blob/5aba4963e3e8ddfc912747076f5b7fe7a743cfe2/src.ts/transaction/transaction.ts#L1004

```js
assert(this.maxFeePerGas >= this.maxPriorityFeePerGas, "priorityFee cannot be more than maxFee", "BAD_DATA", { value: this });
```


bajtos (approved):
👌🏻

PR #347, opened 2024-09-05
feat: record stats for majority vs. committee size ratio

bajtos:
I'd like to get visibility into the size of the majorities - is it a tight majority or a super majority?

Links:

### DIFF
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -196,7 +196,7 @@ export const recordCommitteeSizes = (committees, point) => {
    * participants: Set<string>;
    * nodes: Set<string>;
    * measurements: number;
-   * majoritySize: number;
+   * majoritySize: number | undefined;
    * }>} */
   const stats = new Map()
   for (const c of committees) {
@@ -208,7 +208,7 @@ export const recordCommitteeSizes = (committees, point) => {
         participants: new Set(),
         nodes: new Set(),
         measurements: 0,
-        majoritySize: 0
+        majoritySize: undefined
       }
       stats.set(key, data)
     }
@@ -220,7 +220,9 @@ export const recordCommitteeSizes = (committees, point) => {
       data.nodes.add(`${m.inet_group}::${m.participantAddress}`)
       data.measurements++
 
-      if (m.fraudAssessment === 'OK') data.majoritySize++
+      if (m.fraudAssessment === 'OK') {
+        data.majoritySize = (data.majoritySize ?? 0) + 1
+      }
     }
   }
 
@@ -241,7 +243,7 @@ export const recordCommitteeSizes = (committees, point) => {
     measurementCounts.push(measurements)
 
     // Ignore committees that are too small or with no majority
-    if (majoritySize > 0) {
+    if (majoritySize !== undefined) {
       assert(measurements > 0, 'if there is a majority, there have to be measurements')
       majorityToCommitteeRatios.push(Math.floor(100 * majoritySize / measurements))
     }
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -194,7 +194,8 @@ export const recordCommitteeSizes = (committees, point) => {
    * subnets: Set<string>;
    * participants: Set<string>;
    * nodes: Set<string>;
-   * measurements: number
+   * measurements: number;
+   * majoritySize: number;
    * }>} */
   const stats = new Map()
   for (const c of committees) {
@@ -205,7 +206,8 @@ export const recordCommitteeSizes = (committees, point) => {
         subnets: new Set(),
         participants: new Set(),
         nodes: new Set(),
-        measurements: 0
+        measurements: 0,
+        majoritySize: 0
       }
       stats.set(key, data)
     }
@@ -216,6 +218,8 @@ export const recordCommitteeSizes = (committees, point) => {
       // The pair (inet_group, participant_address) is a good approximation.
       data.nodes.add(`${m.inet_group}::${m.participantAddress}`)
       data.measurements++
+
+      if (m.fraudAssessment === 'OK') data.majoritySize++
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/retrieval-stats.js
```diff
@@ -227,15 +231,23 @@ export const recordCommitteeSizes = (committees, point) => {
...
+  /** @type {Array<number>} */
+  const majoritySizes = []
+  for (const { subnets, participants, nodes, measurements, majoritySize } of stats.values()) {
     subnetCounts.push(subnets.size)
     participantCounts.push(participants.size)
     nodeCounts.push(nodes.size)
     measurementCounts.push(measurements)
+
+    // Ignore committees that are too small or with no majority
+    if (majoritySize > 0 && measurements > 0) {
```
juliangruber, line 243:
```suggestion
    // Ignore committees that are too small or with no majority
    if (measurements > 0 && majoritySize > 0) {
```

I think in this order the condition matches the comment

juliangruber: Also, size zero != "too small", right? You could argue that a committee with one measurement is also too small. Is it instead that we want to ignore just empty committees?

lib/retrieval-stats.js
```diff
@@ -227,15 +231,23 @@ export const recordCommitteeSizes = (committees, point) => {
...
+      majoritySizes.push(Math.floor(100 * majoritySize / measurements))
```
juliangruber, line 244: Why are we multiplying by 100? And dividing by measurements? Shouldn't `majoritySizes` contain `majoritySize`? I don't understand this translation

juliangruber (changes requested):

lib/retrieval-stats.js
bajtos: I see how this is confusing, I had to re-read the code to understand what's going on.

`majoritySize > 0` means that 1) the committee is large enough to give us confidence in the majority being honest 2) a majority was found in the measurements.

`measurements > 0` is a safety check to prevent division by zero.

Now that I think about it, when `majoritySize > 0` then `measurements > 0` must be true as well, because "majority" size is always smaller or equal thant "committee size" (number of measurements).

I propose reworking the code as follows:

```js
if (majoritySize > 0) {
  assert(measurements > 0, "majority cannot be larger than the committee size")
  // calculation
}
```

How can I improve the comment and/or the condition to make the intention more clear?

I can change `majoritySize` to `number | undefined` so that committees evaluated to COMMITTEE_TOO_SMALL and MAJORITY_NOT_FOUND will get `majoritySize === undefined`.

WDYT?

lib/retrieval-stats.js
bajtos: Our histogram function requires the values to be integers. This code calculates the relative size of the majority vs. the committee in percentages.

How can I improve the code to make this intent more obvious?

Alternatively, I can modify the histogram function to support both integers and floats, but I am not convinced the extra complexity is worth it.

juliangruber: Isn't `majoritySize` an integer though?

lib/retrieval-stats.js
bajtos: > Sounds good! What about
> 
> ```js
> if (majoritySize > 0) {
>   assert(measurements > 0, "majority cannot be empty")
> }
> ```


That's not what we are asserting.

```
measurements = majoritySize + sumOf(minoritySizes)
```

lib/retrieval-stats.js
bajtos: Yes, it's an integer.

I am not interested in absolute values - a majority of 60 in a committee of 100 is not great, while a majority of 60 in the committee of 70 gives us a lot of confidence in the correctness of the reported result.

I am interested in the relative size of the majority vs. the entire committee. Is it a tight majority (60 out of 100) or a super majority (60 out ot 70)?

We have 1000 committees, it's not practical to record this relative size for every committee, that's why I am recording percentile values.

lib/retrieval-stats.js
juliangruber: Isn't that just what you suggested above, only with a different assertion message?

lib/retrieval-stats.js
juliangruber: Alright, can we change the name then please? We're not recording the majority size, we're recording it relative to something to the measurement count. I misread your intention from the variable names.

What about majorityToCommitteeSize?

lib/retrieval-stats.js
bajtos: It is the same. I find the assertion message you proposed misleading.

We are not asserting that the majority cannot be empty - we already know that inside the `if (majoritySize > 0)` branch.

We are asserting that the invariants hold, i.e. `majoritySize <= measurements`.

Maybe this will be easier to reason about if we rename `measurement` to `committeeSize`. Then your proposal says this:

```
assert(committeeSize > 0, "majority cannot be empty")
```

Do you see now why I find that misleading?

lib/retrieval-stats.js
bajtos: > Alright, can we change the name then please?

💯 

> We're not recording the majority size, we're recording it relative to something to the measurement count. I misread your intention from the variable names.
> 
> What about majorityToCommitteeSize?

We are not recording size, but a fraction expressed in percentages using the 0-100 range.

I guess that leads us to `majorityToCommitteeRatio`.

https://dictionary.cambridge.org/dictionary/english/ratio
> **the ratio of sth to sth** 
> The ratio of exports to imports also improved from 70.4% to 83.2%.

lib/retrieval-stats.js
juliangruber: I see! To me we don't need to rename the variables, only change the assertion message.

With this one I got confused:

```js
if (majoritySize > 0) {
  assert(measurements > 0, "majority cannot be larger than the committee size")
  // calculation
}
```

We're not actually testing whether one is bigger than the other, so it relies on another invariant.

What about we make this as simple as possible:

```js
if (majoritySize > 0) {
  assert(measurements > 0, "if there is a majority, there have to be measurements")
  // calculation
}
```

bajtos:
I added three new changes:

(1)
```js
if (majoritySize > 0) {
  assert(measurements > 0, "if there is a majority, there have to be measurements")
  // calculation
}
```

(2)
Renamed `majoritySizes` to `majorityToCommitteeRatios`

(3)
Renamed the InfluxDB field prefix from `majority_sizes_percents_` to `majority_ratios_percents_`

@juliangruber LGTY now?

lib/retrieval-stats.js
bajtos: @juliangruber would you like me to apply the following change as well?

> I can change `majoritySize` to `number | undefined` so that committees evaluated to COMMITTEE_TOO_SMALL and MAJORITY_NOT_FOUND will get `majoritySize === undefined`.

juliangruber: Sgtm, but I don't have a strong opinion on it

bajtos: See d5c8883

juliangruber (approved):

PR #352, opened 2024-09-10
chore: remove unneeded test-helper

PatrickNercessian:
Link:

https://github.com/filecoin-station/spark-stats/pull/210


bajtos (approved):

juliangruber:
@PatrickNercessian is this still a draft?

PatrickNercessian:
> @PatrickNercessian is this still a draft?

Yes, it should probably be landed after https://github.com/filecoin-station/spark-stats/pull/210 because it will otherwise break spark-stats tests (which rely on this helper).

Technically it wouldn't instantly break it, since we would need to push to spark-stats to use the updated spark-evaluate as a dependency, but it would thus prevent us from updating the spark-evaluate dependency until the other PR landed

Working on making the changes to that other PR now

bajtos:
Landed, thank you @PatrickNercessian!

PR #355, opened 2024-09-11
Add integrate recent participants contract

juliangruber:
https://github.com/filecoin-station/spark-evaluations-recent-participants

Also added a refactor of `ie-contract.js`, while I was there.


lib/contracts.js
```diff
@@ -0,0 +1,30 @@
...
+    SparkImpactEvaluator.ABI,
+    provider
+  )
+}
+
+export const recentParticipantsContract = new ethers.Contract(
+  SparkEvaluationsRecentParticipants.ADDRESS,
+  SparkEvaluationsRecentParticipants.ABI,
+  provider
+)
```
bajtos, line 30: Do you know why we need a factory function for the meridian contract (`createMeridianContract`)? I find it surprising that there isn't a factory function for "recent participants" contract as well.

bajtos (changes requested):
IIRC, we were discussing a different approach in the past:

1. spark-evaluate creates a JSON object linking to measurement commitments included in the evaluation plus the evaluation outcomes
2. spark-evaluate uploads this JSON object to web3storage and commits the CID of this JSON object to the smart contract
3. the smart contract emits an event with the round index and evaluation commitment CID
4. spark-stats listens for these events, downloads the evaluation results & linked measurements, and runs data analysis & aggregation over the data obtained this way

I'd rather avoid creating many smart contract methods & events individually for each stats field like "recent participants" because it would be a lot of work, IMO.

In particular, if we are interested in recent participants only, then I don't think we need a new contract for that - we can find the participant addresses either in the committed measurements or in the history of `setScores` calls.

PR #356, opened 2024-09-12
Add use `NonceManager` to prevent non-serial nonce errors

juliangruber:
This should help prevent errors like

> spark-evaluate CANNOT SUBMIT SCORES FOR ROUND 14990n (CALL 9/9): Error: nonce has already been used 

As suggested in https://github.com/ethers-io/ethers.js/issues/4258#issuecomment-1705709167

### DIFF
--- a/bin/spark-evaluate.js
+++ b/bin/spark-evaluate.js
@@ -30,10 +30,11 @@ await migrateWithPgConfig({ connectionString: DATABASE_URL })
 
 const { ieContract, provider } = await createMeridianContract()
 
-const signer = ethers.Wallet.fromPhrase(WALLET_SEED, provider)
-const walletDelegatedAddress = newDelegatedEthAddress(/** @type {any} */(signer.address), CoinType.MAIN).toString()
+const wallet = ethers.Wallet.fromPhrase(WALLET_SEED, provider)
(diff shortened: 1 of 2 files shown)


bin/spark-evaluate.js
```diff
@@ -30,10 +30,11 @@ await migrateWithPgConfig({ connectionString: DATABASE_URL })
 
 const { ieContract, provider } = await createMeridianContract()
 
-const signer = ethers.Wallet.fromPhrase(WALLET_SEED, provider)
-const walletDelegatedAddress = newDelegatedEthAddress(/** @type {any} */(signer.address), CoinType.MAIN).toString()
+const wallet = ethers.Wallet.fromPhrase(WALLET_SEED, provider)
+const signer = new ethers.NonceManager(wallet)
+const walletDelegatedAddress = newDelegatedEthAddress(/** @type {any} */(wallet.address), CoinType.MAIN).toString()
```
juliangruber, line 35: unfortunately `signer.signer.address` isn't available in types, although in praxis it is

bajtos (approved):
I find it ugly that we have to pass both `address` and `signer` in function parameters, but I gues that's a minor detail we can improve later.

The rest LGTM 👍🏻

PR #363, opened 2024-09-20
add `cancel-stuck-transactions`

juliangruber:


bin/spark-evaluate.js
```diff
@@ -43,6 +48,40 @@ const createPgClient = async () => {
   return pgClient
 }
 
+const pgClient = await createPgClient()
+const cancelStuckTransactions = new CancelStuckTransactions({
```
bajtos, line 52: Can you please move this to a new file in `lib`? I'd like to keep `bin` files only a thin wrappers around stuff imported from elsewhere.

bin/spark-evaluate.js
```diff
@@ -43,6 +48,40 @@ const createPgClient = async () => {
...
+  async store ({ hash, timestamp, from, maxPriorityFeePerGas, nonce }) {
+    await pgClient.query(`
+      INSERT INTO transactions_pending (hash, timestamp, from, max_priority_fee_per_gas, nonce)
+      VALUES ($1, $2, $3, $4, $5)
+      `,
+      [hash, timestamp, from, maxPriorityFeePerGas, nonce]
+    )
```
bajtos, line 59: Is it a good idea to couple the DB schema in spark-evaluate to what `cancel-stuck-transactions` need? If the format of data stored by `cancel-stuck-transactions` changes, we will need to evolve the DB schema too.

Have you considered storing this data as JSON instead?

```js
async store ({hash, ...txData}) {
    await pgClient.query(`
      INSERT INTO transactions_pending (hash, data)
      VALUES ($1)
      `,
      [hash, txData]
    )
}

File: bin/spark-evaluate.js [hunk 3]
```diff
@@ -51,11 +90,18 @@ await Promise.all([
...
+    while (true) {
+      try {
+        await cancelStuckTransactions.olderThan(2 * ROUND_LENGTH_MS)
+      } catch (err) {
+        console.error(err)
+        Sentry.captureException(err)
+      }
+      await timers.setTimeout(CHECK_STUCK_TXS_DELAY)
+    }
+  })()
```
bajtos, line 106: Let's move this to a named function please, that function should be exported by the same file that exports the main implementation (see one of my comments above).

lib/evaluate.js
```diff
@@ -179,6 +183,11 @@ export const evaluate = async ({
   }
   const setScoresDuration = Date.now() - start
 
+  ;(async () => {
+    await tx.wait()
+    cancelStuckTransactions.successful(tx)
+  })().catch(console.error)
```
bajtos, line 189: Since you are already using `.catch`, how about this:

```js
  // run in the background
  tx.wait()
    .then(() => cancelStuckTransactions.successful(tx))
    .catch(console.error)
```

bajtos (approved):
I am concerned about coupling the database schema with the data structure used by the cancellation library. But I guess that's easy to change later.

The rest of the changes LGTM.

Remember to fix the failing tests.

bin/spark-evaluate.js
juliangruber: I don't consider them coupled - `cancel-stuck-transactions` is a versioned library and if it changes the schema then we will get a dependabot PR and failing CI. I think it's great to fully SQL this, because it means we can add monitoring, plus it adds just a little bit of type safety.

If you feel strongly I can change it to JSON however, I'm not married to this method.

lib/evaluate.js
juliangruber: I'm using the wrapper with catch here because it's a pattern:

```js
;(async () => {
  // everything in here happens outside the flow
}).catch(() => {
  // we always need error handling with this pattern
})
```

I try to avoid `Promise#then()` and `Promise#catch()` as much as possible, because they can be seen as lower-level building blocks that should be replaced by `async/await` and `try/catch` as much as possible

juliangruber:
Converting to draft because
- type tests are failing
- there are merge conflicts

bin/spark-evaluate.js
bajtos: > I don't consider them coupled - `cancel-stuck-transactions` is a versioned library and if it changes the schema then we will get a dependabot PR and failing CI. I think it's great to fully SQL this, because it means we can add monitoring, plus it adds just a little bit of type safety.

You convinced me 👍🏻

lib/cancel-stuck-transactions.js
```diff
@@ -0,0 +1,66 @@
...
+      },
+      async list () {
+        const { rows } = await pgClient.query('SELECT * FROM transactions_pending')
+        return rows.map(row => ({
+          hash: row.hash,
+          timestamp: row.timestamp,
+          from: row.from_address,
+          maxPriorityFeePerGas: row.max_priority_fee_per_gas,
+          nonce: row.nonce
+        }))
```
bajtos, line 28: Nitpick, feel free to ignore.

If the only reason for this `rows.map` function is to rename the columns, then I think the following would be slightly better, as the database will complain you reference a column that does not exist. Notice that I am enclosing the final column names in double quotes to prevent PG from converting them to lowercase (plus `from` is likely a reserved word?).

```suggestion
        const { rows } = await pgClient.query(`
          SELECT 
            hash, 
            timestamp, 
            from_address as "from",
            max_priority_fee_per_gas as "maxPriorityFeePerGas",
            nonce
          FROM transactions_pending
        `)
        return rows
```

lib/evaluate.js
```diff
@@ -147,9 +149,10 @@ export const evaluate = async ({
 
   const start = Date.now()
   const buckets = createSetScoresBuckets(participants)
+  let tx
   for (const [bucketIndex, bucket] of Object.entries(buckets)) {
     try {
```
bajtos, line 154: Here is how I understand your new version at the high level:


```js
  let tx
  for (const [bucketIndex, bucket] of Object.entries(buckets)) {
    tx = await ieContractWithSigner.setScores // etc.
  }
  tx.wait()
```

This code will wait only for the last transaction created.

If that's intentional, then let's make it more obvious to the reader, please.

For example:

```js
  let lastTx
  for (const [bucketIndex, bucket] of Object.entries(buckets)) {
    const tx = await ieContractWithSigner.setScores // etc.
    lastTx = tx
  }
  // add a comment explaining why we wait for the last transaction only
  // and don't need to wait for other transactions
  lastTx.wait()
```

lib/evaluate.js
```diff
@@ -179,6 +183,11 @@ export const evaluate = async ({
   }
   const setScoresDuration = Date.now() - start
 
+  ;(async () => {
+    await tx.wait()
+    stuckTransactionsCanceller.removeConfirmed(tx)
```
bajtos, line 188: IIUC (see my comment above), you are going to call `removeConfirmed` only the last transaction. That does not look correct to me. IMO, we need to remove all confirmed transactions created in this round.

migrations/017.do.cancel-stuck-transactions.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE transactions_pending (
+  hash TEXT NOT NULL PRIMARY KEY,
+  timestamp TEXT NOT NULL,
```
bajtos, line 3:
```suggestion
  timestamp TIMESTAMPTZ NOT NULL,
```

Since our goal is to have more type safety.

bajtos (changes requested):

lib/cancel-stuck-transactions.js
juliangruber: good idea!

lib/evaluate.js
juliangruber: This was even more broken. Each tx should be awaited and then marked as confirmed. Fixed 👍

bajtos (approved):
👌🏻

lib/evaluate.js
bajtos: It may be worth creating a test to verify this works as intended. However, mocking all the dependencies to simulate two TXs running concurrently will require quite a bit of code, so it may not be worth it right now.

juliangruber:
Will now mirror this to `spark-publish`

juliangruber:
> ## Suspect Issues
> 
> * ‼️ **TypeError: Cannot mix BigInt and other types, use explicit conversions** `startCancelStuckTransactions(cancel-stuck-trans...` [View Issue](https://space-meridian.sentry.io/issues/5903978203/?referrer=github-pr-bot)

Strange! In the example script of `cancel-stuck-transaction`, it works, with the same version of `ethers`. I'm taking a closer look.

> * ‼️ **Error: Filfox request failed with 500: {"statusCode":500,"message":"Internal server error"}** `startCancelStuckTransactions(cancel-stuck-trans...` [View Issue](https://space-meridian.sentry.io/issues/5904426486/?referrer=github-pr-bot)

I'll either add retry logic or let it fail more silently (the whole op will be retried anyway)

> 

juliangruber:
> Strange! In the example script of cancel-stuck-transaction, it works, with the same version of ethers. I'm taking a closer look.

I think I know what happens: When reading from mysql, it doesn't return a BigInt

juliangruber:
https://github.com/brianc/node-pg-types/issues/78

juliangruber:
https://github.com/filecoin-station/spark-evaluate/pull/368

juliangruber:
- [ ] https://github.com/filecoin-station/cancel-stuck-transactions/pull/2
- [ ] https://github.com/filecoin-station/spark-evaluate/pull/369

PR #368, opened 2024-09-25
cancel-stuck-transactions: fix return bigint type

juliangruber:
Fixes https://space-meridian.sentry.io/issues/5903978203/?project=4505906069766144&referrer=github-pr-bot

### DIFF
--- a/lib/cancel-stuck-transactions.js
+++ b/lib/cancel-stuck-transactions.js
@@ -27,7 +27,10 @@ export const createStuckTransactionsCanceller = ({ pgClient, signer }) => {
             nonce
           FROM transactions_pending
         `)
-        return rows
+        return rows.map(row => ({
+          ...row,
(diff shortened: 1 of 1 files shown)


bajtos (approved):
I feel this bug is telling us that we need to add tests for the "store" component. Since it requires a bit of refactoring to extract the store into something that can be tested independently from the TX cancellation, I am fine with landing this ASAP as an urgent fix, as long as the test are added very soon in a follow-up PR.

juliangruber:
I don't have the motivation for writing this test, especially this week, so I can't merge this PR now. Could you please take this over then?

To me the risk is low. The stuck tx cancelling logic happens out of band and when it fails is easy to fix.

bajtos:
Fair enough. Let's land this PR now and defer writing tests until the next time. 🤷🏻

bajtos:
I landed the PR to prevent us from exhausting the error spend allocation in Sentry.

Screenshot 2024-09-25 at 14 38 37

PR #373, opened 2024-10-01
feat: stop populating daily_stations table

bajtos:
This is the only time-sensitive task remaining in [Fix the performance of the Station Public Dashboard](https://github.com/space-meridian/roadmap/issues/132). The change prevents our PG DB from consuming too much disk space.

I ran `rg daily_stations` in https://github.com/filecoin-station/spark-stats and we are no longer querying that table, so it's safe to stop writing new data.

I am keeping the table around, so that we can use the historical data stored in this table to backfill new tables we created as part of the fix.

Links:
- https://github.com/space-meridian/roadmap/issues/132

### DIFF
--- a/lib/platform-stats.js
+++ b/lib/platform-stats.js
@@ -7,96 +7,12 @@ const debug = createDebug('spark:platform-stats')
 
 /**
  * @param {import('pg').Client} pgClient
- * @param {import('./preprocess.js').Measurement[]} honestMeasurements
  * @param {import('./preprocess.js').Measurement[]} allMeasurements
  */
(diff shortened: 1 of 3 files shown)


juliangruber:
> I ran `rg daily_stations` in https://github.com/filecoin-station/spark-stats and we are no longer querying that table, so it's safe to stop writing new data.

Btw a safer way to search for use is https://github.com/search?q=org%3Afilecoin-station%20daily_stations&type=code

juliangruber (approved):
Feels good!

PR #378, opened 2024-10-09
fix cancel stuck transactions when previous run not yet finished

juliangruber:
Otherwise while we're stuck waiting for replacement transactions to be confirmed, we're not trying to cancel any new stuck transactions.


bajtos (approved):
I have a gut feeling that this new way may schedule too many replacement transactions, but I don't have the capacity now to think that through.

Other than that, the change looks good to me.

PR #383, opened 2024-10-20
Add publish rsr to smart contract

juliangruber:
- [x] Deploy rsr contract to mainnet
- [x] Publish rsr
- [x] Publish own git hash
- [x] Write tests
- [x] Create Storacha space
- [x] Configure Storacha environment variables
- [x] Update schema
- [x] Publish as IPLD
- [x] Publish round details to Storacha
- [ ] Dynamically split CAR into multiple blocks (should not be blocking release)


lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+        providers.set(minerId, { successful: 0, total: 0 })
+      }
+      providers.get(minerId).successful += successful
+      providers.get(minerId).total += total
+    }
+  }
+
+  const directoryCid = await pRetry(() => storachaClient.uploadDirectory([
+    new File([JSON.stringify({
+      date: rows[0].evaluated_at.toISOString().split('T')[0],
```
bajtos, line 59: Do you remember our discussion about the difficulties of converting JavaScript timestamps to dates in a correct way that takes into account timezones?
https://github.com/filecoin-station/spark-stats/pull/210#discussion_r1751821906

Let's avoid it by performing the "timestamp to day" conversion in the database.

```
SELECT 
  evaluated_at::DATE,
  -- etc.
FROM publish_rsr_rounds
-- etc.
```

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+      return acc
+    }, {})
+  ])
+})
+
+const publishRsr = withPgClient(async ({ pgClient, storachaClient, rsrContract }) => {
+  const { rows } = await pgClient.query(`
+    SELECT *
+    FROM publish_rsr_rounds
+    WHERE evaluated_at <= now()::date AND evaluated_at > now()::date - interval '1 day'
```
bajtos, line 42: Is it safe to publish commitments for the current day, before the day is over? What if we end up adding one more evaluation after this code was executed?

I propose to select data from yesterday - that data is not going to change.

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+  try {
+    return await fn({ pgClient, ...args })
+  } finally {
+    await pgClient.end()
+  }
+}
+
+export const postEvaluate = withPgClient(async ({ pgClient, round, gitCommit }) => {
+  await pgClient.query(`
+    INSERT INTO publish_rsr_rounds
```
bajtos, line 18: We usually start variable/property/table names with a noun, and function/method names with a verb.

Let's rename `publish_rsr_rounds` to follow that convention.

For example: `unpublished_rsr_rounds`, `evaluated_rsr_round`, `recent_rsr_rounds`.

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+
+  const directoryCid = await pRetry(() => storachaClient.uploadDirectory([
+    new File([JSON.stringify({
+      date: rows[0].evaluated_at.toISOString().split('T')[0],
+      meta: {
+        rounds: Object.fromEntries(rows.map(row => [
+          row.round_index,
+          {
+            sparkEvaluateVersion: row.spark_evaluate_version,
+            measurementCommitments: row.measurement_commitments,
```
bajtos, line 65: Should we use IPLD links here?

```
measurementCommitments: row.measurement_commitments.map(cid => {['/']: cid})
```

Or perhaps serialise the data using https://www.npmjs.com/package/@ipld/dag-json, I think that should handle CID links.

This is not strictly required; it depends on how much we want to be compatible with the rest of the IPLD tooling.

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+  ]))
+  console.log(`https://${directoryCid}.ipfs.w3s.link/commitment.json`)
+
+  const tx = await pRetry(() => rsrContract.addCommitment(directoryCid.toString()))
+  console.log(tx.hash)
+  await tx.wait()
+
+  await pgClient.query(`
+    DELETE FROM publish_rsr_rounds
+    WHERE round_index < $1
```
bajtos, line 81: At the top, you select only the following rows for publishing:

```
WHERE evaluated_at <= now()::date AND evaluated_at > now()::date - interval '1 day'
```

Here, you are deleting all rows with `round_index` before `rows[0].round_index`, excluding `rows[0].round_index`.

1. What happens if we cannot publish measurements for one day, and when the next day comes, we manage to publish measurements for that new day. I believe we will lose the measurements for that first day forever. Is that fine?
2. If you want to delete rows by `round_index`, shouldn't you delete all rows where round_index is _less or equal_ to the _highest_ `round_index` in the committed data?

Please consider starting with a failing test before addressing each of these two cases.

bajtos (changes requested):
Great start!

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+    round.cids,
+    round.details,
+    round.measurements.reduce((acc, m) => {
+      acc[m.minerId] = acc[m.minerId] || { successful: 0, total: 0 }
+      acc[m.minerId].total++
+      if (m.retrievalResult === 'OK') {
+        acc[m.minerId].successful++
+      }
+      return acc
+    }, {})
```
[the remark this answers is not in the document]
juliangruber: The on-chain data shouldn't be based on committees, is your suggestion still true?

bin/spark-evaluate.js
```diff
@@ -41,12 +66,20 @@ const createPgClient = async () => {
...
+await Promise.all([
+  startEvaluate({
+    ieContract,
+    fetchMeasurements,
+    fetchRoundDetails,
+    recordTelemetry,
+    createPgClient,
+    logger: console,
+    setScores: (participants, values) => setScores(signer, participants, values),
+    gitCommit: GIT_COMMIT
```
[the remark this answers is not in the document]
juliangruber: Great idea!

lib/publish-rsr.js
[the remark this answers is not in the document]
juliangruber: Since we want this to be the per-measurement RSR, not per-committee. But I think indeed I'm missing the IPv4 logic, and potentially other parts, as `.total` should only be accepted measurements. I believe we need this:

```js
let total = 0
let successful = 0
  if (m.fraudAssessment === 'ok') total++
  if (m.retrievalResult === 'ok') successful++
}
```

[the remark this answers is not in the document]
juliangruber: https://github.com/filecoin-station/spark-evaluate/pull/383/commits/f5e6ee0019f372f2e2fbda5c70ae5f9c0a09e020

lib/publish-rsr.js
juliangruber: If we do that, pg will also return a date object:

```js
const { rows: [{ now }] } = await pgClient.query(`
  SELECT now()::DATE
`)
console.log({ now, type: typeof now })
```

```js
{ now: 2024-10-23T22:00:00.000Z, type: 'object' }
```

Do you know how else to achieve this?

lib/publish-rsr.js
juliangruber: `now()::date` is 0am today, so this is querying from yesterday's 0am to today's 0am, which is what you're suggesting. Right?

lib/publish-rsr.js
juliangruber: https://space-meridian.slack.com/archives/C07CPEGCTFT/p1729776103560939 (private slack)

lib/publish-rsr.js
```diff
@@ -0,0 +1,96 @@
...
+const withPgClient = fn => async ({ createPgClient, ...args }) => {
+  const pgClient = await createPgClient()
+  try {
+    return await fn({ pgClient, ...args })
+  } finally {
+    await pgClient.end()
+  }
+}
+
+export const postEvaluate = withPgClient(async ({ pgClient, round, gitCommit }) => {
```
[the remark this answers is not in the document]
bajtos: FWIW, the contract API uses `providerRetrievalResultStatsCommitment` now.

lib/publish-rsr.js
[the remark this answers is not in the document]
bajtos: You are excluding minority results from RSR.

See https://github.com/filecoin-station/spark-evaluate/commit/3b31a57a483fba5d61e1b280b695dcf1ba73d04e

> This commit fixes the problem by changing the implementation of
`updatePublicStats` to iterate over all measurements assigned to all
committees. This way we include all measurements in the calculation:
minority measurements, measurements where no majority was found,
measurements belonging to committees that are too small.

Is it okay to have a different per-measurement RSR in the on-chain commitment than the per-measurement RSR reported by Spark Stats API? If not, then I urge you to share the code building this metric.

lib/publish-rsr.js
bajtos: How about this? 

```sql
SELECT now()::DATE::TEXT
```

lib/publish-rsr.js
bajtos: > `now()::date` is 0am today, so this is querying from yesterday's 0am to today's 0am, which is what you're suggesting. Right?

I see! That's clever 👍🏻

Can we find a way how to make this condition easier to understand?

We want `evaluated_at` to be in the interval `now()::date - interval '1 day' to `now()::date`.

I would also include 0:00am in the same day, not in the previous day as you do know. It's unlikely this is going to make any difference in the result, but it could make the code easier to understand.

This is how I would write it myself:

```suggestion
    WHERE now()::date - interval '1 <= evaluated_at AND evaluated_at < now()::date 
```

Feel free to keep your current version if you don't see my proposal as a meaningful improvement.

lib/publish-rsr.js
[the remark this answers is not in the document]
bajtos: Let me clarify my previous comment.

In this proposal:

```js
let total = 0
let successful = 0
  if (m.fraudAssessment === 'ok') total++
  if (m.retrievalResult === 'ok') successful++
}
```

- `total` counts only accepted measurements in the majority
- `retrievalResult` counts all measurements reporting successfull retrieval, regardless of whether they were accepted/rejected and in minority/majority

The source of the confusion is that `fraudAssessment` combines two steps - 1) whether the measurement is accepted based on tasking, IPv4 subnets, etc. and 2) whether the measurement is in the majority. This is good for deciding which measurements to reward - only those with `ok` assessment, but it makes RSR calculation more complex.

I think it would be worth taking a closer look at this and seeing how to refactor the code to make these two evaluation step outcomes more explicit & distinguished. But that's out of the scope of this PR.

migrations/020.do.publish-rsr.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE unpublished_rsr_rounds (
+  round_index NUMERIC PRIMARY KEY,
+  evaluated_at TIMESTAMPTZ NOT NULL DEFAULT NOW(),
+  spark_evaluate_version TEXT NOT NULL,
+  measurement_commitments TEXT[] NOT NULL,
+  round_details JSONB NOT NULL,
+  providers JSONB NOT NULL
```
bajtos, line 7: I find the name `providers` as too ambiguous/potentially confusing, since this does not store a list of providers, but a list of per-provider retrieval stats.

The smart-contract uses the name `providerRetrievalResultStats`, which I find too long and redundant in the scope of `unpublished_rsr_rounds`.

How about this?

```suggestion
  providers_stats JSONB NOT NULL
```

Not a big deal, we can keep the current name if you prefer. It's an internal implementation detail that can be easily changed later.

migrations/020.do.publish-rsr.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE unpublished_rsr_rounds (
+  round_index NUMERIC PRIMARY KEY,
```
bajtos, line 2: ❗ 

I realised one important thing: the round index is local to the Meridian smart contract version. When we deploy a new Meridian version, round index resets back to zero.

I propose to add the smart contract address to the per-round data inside the evaluation commitment.

In other words, round id should be a tuple `(contractAddress, roundIndex)`.

This affects the committed JSON structure too: Instead of converting this tuple to a string id that can be used as a key (property name), I think it will be better to change “rounds” from an object to an array, as was proposed in the call yesterday (or on Slack?).

Notion discussion:
https://www.notion.so/Design-Spark-SLI-on-chain-interface-121837df73d4801aaaa4f13ff7839cd0?d=b58a1a22a5d248258cae0bbb5c4d9620&pvs=4#b70f9a9b99dd43739b136393d08c2091

Here is the schema we use in spark-api's table `spark_rounds` - see `meridian_address` and `meridian_round`.

```
spark=# \d+ spark_rounds;
                                                     Table "public.spark_rounds"
       Column       |           Type           | Collation | Nullable | Default 
--------------------+--------------------------+-----------+----------+---------
 id                 | bigint                   |           | not null |         
 created_at         | timestamp with time zone |           | not null |         
 meridian_address   | text                     |           |          |         
 meridian_round     | bigint                   |           |          |         
 max_tasks_per_node | integer                  |           | not null | 360     
 start_epoch        | bigint                   |           | not null |         
 measurement_count  | bigint                   |           |          |         
```

The column `id` contains the round number that we create internally inside spark-api by tracking smart contract versions and mapping per-contract-round-index to this round number. It seemed like a good idea when I started building Spark. I discovered the limitations of that approach very soon and switched to `(meridian_address, meridian_rond)` in most places.

Dockerfile
```diff
@@ -14,6 +14,9 @@ ENV NODE_ENV=production
 # 4096MB available memory - 200MB for anything else
 ENV NODE_OPTIONS="--max-old-space-size=3896"
 
+ARG GIT_COMMIT
+ENV GIT_COMMIT=$GIT_COMMIT
```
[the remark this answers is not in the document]
juliangruber: Got it! I'm happy to document it, where's a good place for that?

lib/publish-rsr.js
juliangruber: > What happens if we cannot publish measurements for one day, and when the next day comes, we manage to publish measurements for that new day. I believe we will lose the measurements for that first day forever. Is that fine?

It would be better to recover from this, and keep publishing until the buffer is empty. Will fix.

> If you want to delete rows by round_index, shouldn't you delete all rows where round_index is less or equal to the highest round_index in the committed data?

You're right! Fixed to

```sql
await pgClient.query(`
  DELETE FROM unpublished_rsr_rounds WHERE round_index <= $1
`, [rows.at(-1).round_index])
```

lib/publish-rsr.js
juliangruber: It's IPLD serialized as dag-json now

juliangruber:
@bajtos hindsight please

PR #385, opened 2024-10-22
feat: distinguish HTTP status from Lassie status

bajtos:
Remove the following result codes:
- `ERROR_${number}`
- `BAD_GATEWAY`
- `GATEWAY_TIMEOUT`

Replace them with HTTP status codes prefixed with `HTTP_` if the
retrieval was performed directly over Trustless HTTP GW protocol
and `LASSIE_` if the retrieval was performed using Lassie. In the latter
case, the status code comes from Lassie, not the storage provider.

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -250,7 +250,7 @@ const assertValidMeasurement = measurement => {
  * @param {Partial<import('./typings.js').RawMeasurement>} measurement
  * @return {import('./typings.js').RetrievalResult}
  */
-export const getRetrievalResult = measurement => {
+export const getRetrievalResult = (measurement) => {
   switch (measurement.indexer_result) {
     case 'OK':
     case 'HTTP_NOT_ADVERTISED':
@@ -270,8 +270,6 @@ export const getRetrievalResult = measurement => {
   }
 
   switch (measurement.status_code) {
-    case 502: return 'BAD_GATEWAY'
-    case 504: return 'GATEWAY_TIMEOUT'
     case 600: return 'UNKNOWN_FETCH_ERROR'
     case 801: return 'HOSTNAME_DNS_ERROR'
     case 802: return 'CONNECTION_REFUSED'
@@ -281,7 +279,11 @@ export const getRetrievalResult = measurement => {
     case 904: return 'CANNOT_PARSE_CAR_FILE'
   }
 
-  if (measurement.status_code >= 300) return `ERROR_${measurement.status_code}`
+  if (measurement.status_code >= 300) {
+    const prefix = measurement.protocol === 'http' ? 'HTTP_' : 'LASSIE_'
+    // I cannot use + concatenation because TypeScript would complain
+    return `${prefix}${measurement.status_code}`
+  }
 
   const ok = measurement.status_code >= 200 && typeof measurement.end_at === 'string'
 
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -39,10 +39,7 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   /** @type {Partial<Record<import('./typings.js').RetrievalResult, number>>} */
   const resultBreakdown = {
     OK: 0,
-    TIMEOUT: 0,
-    CAR_TOO_LARGE: 0,
-    BAD_GATEWAY: 0,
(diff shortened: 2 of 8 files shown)


lib/preprocess.js
```diff
@@ -281,7 +279,11 @@ export const getRetrievalResult = measurement => {
     case 904: return 'CANNOT_PARSE_CAR_FILE'
   }
 
-  if (measurement.status_code >= 300) return `ERROR_${measurement.status_code}`
+  if (measurement.status_code >= 300) {
+    const prefix = measurement.protocol === 'http' ? 'HTTP_' : 'LASSIE_'
+    // I cannot use + concatenation because TypeScript would complain
```
juliangruber, line 284:
```suggestion
```

String interpolation is the idiomatic JS way now anyway

lib/preprocess.js
```diff
@@ -281,7 +279,11 @@ export const getRetrievalResult = measurement => {
...
+    return `${prefix}${measurement.status_code}`
```
juliangruber, line 285:
```suggestion
    return `${prefix}ERROR_${measurement.status_code}`
```

Shall we keep `ERROR_` to clarify the meaning? Or is this always an error anyway?

lib/retrieval-stats.js
```diff
@@ -39,10 +39,7 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   /** @type {Partial<Record<import('./typings.js').RetrievalResult, number>>} */
   const resultBreakdown = {
     OK: 0,
-    TIMEOUT: 0,
-    CAR_TOO_LARGE: 0,
-    BAD_GATEWAY: 0,
-    GATEWAY_TIMEOUT: 0
```
juliangruber, line 45: why does this PR remove these 3 result codes from this test?

juliangruber: Shouldn't there be new `LASSIE_` or `HTTP_` codes?

test/helpers/test-data.js
```diff
@@ -20,7 +20,7 @@ export const VALID_MEASUREMENT = {
   provider_address: '/dns4/production-ipfs-peer.pinata.cloud/tcp/3000/ws/p2p/Qma8ddFEQWEU8ijWvdxXm3nxU7oHsRtCykAaVz8WUYhiKn',
   providerId: 'PROVIDERID',
   spark_version: '1.10.4',
-  protocol: 'bitswap',
+  protocol: 'http',
```
juliangruber, line 23: what is this change for?

juliangruber (changes requested):
+1 on the direction!

lib/preprocess.js
bajtos: I guess it depends on how you define an error.

This function returns a "retrieval result code".

There is only one code for a retrieval that was successful - `OK`.

All other codes describe various failure modes.

That's why I think we don't need to include `ERROR` in those codes.

test/helpers/test-data.js
bajtos: It's a bit unrelated. I thought this change would reduce the number of tests I need to fix as part of this PR, but after I reverted the change, all tests are passing, so it's not necessary at all.

Reverted.

lib/retrieval-stats.js
bajtos: See the comment above this object initialisation (GitHub diff does not show it by default, unfortunately):

```js
  // We are intentionally not initializing all possible keys here.
  // Example of omitted keys: UNDEFINED, ERROR_500 and ERROR_404.
  // The idea is that if we don't explicitly initialise them here and there is no measurement with
  // such retrieval result, then the Grafana dashboard will not show these results at all.
```

IIRC, I added a subset of keys to ensure they always show at the top of the legend in Grafana, and that we show these keys (retrieval result codes) even if no measurement reported the key.

Screenshot 2024-10-23 at 13 31 05

I feel this is no longer needed/useful for all of these codes.

On the second thought, maybe `TIMEOUT` is worth keeping.

WDYT?

bajtos: >ensure they always show at the top of the legend in Grafana

This is no longer relevant. We have a transformation in Grafana to (re)order the codes manually.

juliangruber: got it!

juliangruber (approved):

PR #387, opened 2024-10-23
fix: tighten up indexer result enum

bajtos:
Change the `string` type to an enum with the following variants:

    | 'OK'
    | 'HTTP_NOT_ADVERTISED'
    | 'NO_VALID_ADVERTISEMENT'
    | 'ERROR_FETCH'
    | `ERROR_${number}`
    | undefined
    | null;

### DIFF
--- a/test/public-stats.test.js
+++ b/test/public-stats.test.js
@@ -200,9 +200,9 @@ describe('public-stats', () => {
 
       // Notice: this measurement is for the same task as honestMeasurements[0], therefore it's
       // effectively ignored as the other measurement was successful.
-      honestMeasurements.push({ ...VALID_MEASUREMENT, indexerResult: 'UNKNOWN_ERROR' })
+      honestMeasurements.push({ ...VALID_MEASUREMENT, indexerResult: 'ERROR_FETCH' })
       // This is a measurement for a new task.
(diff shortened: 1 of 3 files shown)


test/public-stats.test.js
```diff
@@ -228,7 +228,7 @@ describe('public-stats', () => {
       const honestMeasurements = [
         { ...VALID_MEASUREMENT },
         // HTTP_NOT_ADVERTISED means the deal is indexed
-        { ...VALID_MEASUREMENT, cid: 'bafy2', indexerResult: 'HTTP_NOT_ADVERTISED', retrievalResult: 'IPNI_HTTP_NOT_ADVERTISED' },
+        { ...VALID_MEASUREMENT, cid: 'bafy2', indexerResult: 'HTTP_NOT_ADVERTISED', retrievalResult: 'ERROR_502' },
```
bajtos, line 231: `IPNI_HTTP_NOT_ADVERTISED` is not a valid retrieval result code. The indexer result `HTTP_NOT_ADVERTISED ` means that only Graphsync was advertised, and the retrieval check continues to retrieve the CAR bytes.

juliangruber (approved):

PR #396, opened 2024-11-05
fix: fetch-recent-miner-measurements & committees

bajtos:
- include non-majority results in the output
- optionally, print the committee evaluation result in the TXT summary

### DIFF
--- a/bin/fetch-recent-miner-measurements.js
+++ b/bin/fetch-recent-miner-measurements.js
@@ -16,8 +16,11 @@ import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
 import { evaluate } from '../lib/evaluate.js'
 import { fetchRoundDetails } from '../lib/spark-api.js'
 
+/** @import {Measurement} from '../lib/preprocess.js' */
+
 const {
   STORE_ALL_MINERS,
+  SHOW_EVALUATION_RESULT,
   KEEP_REJECTED
 } = process.env
 
@@ -86,6 +89,7 @@ const MINER_DATA_FILE = `measurements-${minerId}.ndjson`
 const MINER_SUMMARY_FILE = `measurements-${minerId}.txt`
 
 const keepRejected = isFlagEnabled(KEEP_REJECTED)
+const showEvaluationResult = isFlagEnabled(SHOW_EVALUATION_RESULT) || keepRejected
 
 const allMeasurementsWriter = isFlagEnabled(STORE_ALL_MINERS)
   ? fs.createWriteStream(ALL_MEASUREMENTS_FILE)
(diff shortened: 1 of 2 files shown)


bin/fetch-recent-miner-measurements.js
```diff
@@ -218,18 +222,21 @@ async function processRound (roundIndex, measurementCids, resultCounts) {
     prepareProviderRetrievalResultStats: async () => {}
   })
 
+  const isAccepted = (/** @type {Measurement} */ m) =>
```
juliangruber, line 225: Can this be shared with other parts of the repository? Or is this check only useful in this place?

bin/fetch-recent-miner-measurements.js
```diff
@@ -291,17 +298,23 @@ function recordTelemetry (measurementName, fn) {
...
 
-  if (includeFraudAssesment) {
-    fields.push((m.fraudAssessment === 'OK' ? '🫡  ' : '🙅  '))
+  if (showEvaluationResult) {
+    switch (m.fraudAssessment) {
+      case 'OK': fields.push('🫡  '); break
+      case 'MINORITY_RESULT': fields.push('🫠  '); break
+      case 'COMMITTEE_TOO_SMALL': fields.push('🤷 '); break
+      case 'MAJORITY_NOT_FOUND': fields.push('🤷 '); break
+      default: fields.push('🙅 ')
```
juliangruber, line 316: nice use of emoji :D

lib/round.js
```diff
@@ -7,8 +9,9 @@ export class RoundData {
    */
   constructor (index) {
     this.index = index
-    /** @type {import('./preprocess.js').Measurement[]} */
+    /** @type {string[]} */
     this.measurementBatches = []
+    /** @type {Measurement[]} */
```
juliangruber, line 14: Surprising to me that the type checker didn't catch this mistake

juliangruber (approved):

lib/round.js
bajtos: Before this change, the type was `any` - essentially there was no typechecking for this property.

bin/fetch-recent-miner-measurements.js
bajtos: I was thinking about this some more and came to the conclusion that it was a mistake to add committee-based evaluation outcomes to the "fraudAssessment" enum. What I would like to do instead is to have two fields in each measurement:
- `fraudAssessment` based on the tasking algorithm - we can rename it to `taskingEvaluation`
-  `majorityEvaluation` based on the committee & majority algorithm
- rework the code calculating participant scores to look at both `fraudAssessment` and `taskingEvaluation`.

That would prevent mistakes like the one I am fixing in this pull request.

juliangruber: That's a great idea 👍

juliangruber: That sounds great to me!

bajtos: > validationResult - e.g. IPNI_NOT_QUERIED, this can be done in the preprocessing step.

See https://github.com/filecoin-station/spark-evaluate/pull/439

bajtos: See https://github.com/filecoin-station/spark-evaluate/pull/442

PR #401, opened 2024-11-12
cli: split `fetch-recent-miner-measurements` and `evaluate-measurements`

juliangruber:
Usage:

```console
$ node bin/fetch-recent-miner-measurements.js f0...
...
$ node bin/evaluate-measurements measurements-f0....ndjson
 → evaluating round 19117n
 → added 183 accepted measurements from this round
 → evaluating round 19118n
 → added 0 accepted measurements from this round
 → evaluating round 19127n
 → added 181 accepted measurements from this round
 → evaluating round 19128n
 → added 0 accepted measurements from this round
Found 364 accepted measurements.
  IPNI_ERROR_404                           364        (100%)
Wrote evaluation to evaluation-5bf6db8.txt
```

This is a first step towards evaluating arbitrary rounds. Next: https://github.com/filecoin-station/spark-evaluate/pull/402

__After merging, update Spark troubleshooting docs__

### DIFF
--- a/bin/evaluate-measurements.js
+++ b/bin/evaluate-measurements.js
@@ -36,10 +36,13 @@ for (const line of measurementsFile.split('\n').filter(Boolean)) {
   rounds.get(roundIndex).push(measurement)
 }
 
-const EVALUATION_FILE = `${basename(measurementsPath, '.ndjson')}.evaluation.txt`
-const evaluationWriter = fs.createWriteStream(EVALUATION_FILE)
+const EVALUATION_TXT_FILE = `${basename(measurementsPath, '.ndjson')}.evaluation.txt`
+const EVALUATION_NDJSON_FILE = `${basename(measurementsPath, '.ndjson')}.evaluation.ndjson`
 
-evaluationWriter.write(formatHeader({ includeFraudAssesment: keepRejected }) + '\n')
+const evaluationTxtWriter = fs.createWriteStream(EVALUATION_TXT_FILE)
+const evaluationNdjsonWriter = fs.createWriteStream(EVALUATION_NDJSON_FILE)
+
+evaluationTxtWriter.write(formatHeader({ includeFraudAssesment: keepRejected }) + '\n')
 
 const resultCounts = {
   total: 0
@@ -63,7 +66,8 @@ for (const [r, c] of Object.entries(resultCounts)) {
   )
 }
 
-console.error('Wrote evaluation to %s', EVALUATION_FILE)
+console.error('Wrote   human-readable evaluation to %s', EVALUATION_TXT_FILE)
+console.error('Wrote machine-readable evaluation to %s', EVALUATION_NDJSON_FILE)
 
 /**
  * @param {bigint} roundIndex
@@ -107,9 +111,14 @@ async function processRound (roundIndex, measurements, resultCounts) {
       .map(m => ({ ...m, fraudAssessment: undefined }))
   }
 
-  evaluationWriter.write(
+  evaluationTxtWriter.write(
     round.measurements
       .map(m => formatMeasurement(m, { includeFraudAssesment: keepRejected }) + '\n')
+      .join('')
+  )
+  evaluationNdjsonWriter.write(
+    round.measurements
+      .map(m => JSON.stringify(m) + '\n')
       .join('')
   )
   console.error(' → added %s accepted measurements from this round', round.measurements.length)
--- a/bin/evaluate-measurements.js
+++ b/bin/evaluate-measurements.js
@@ -1,0 +1,176 @@
+import fs from 'node:fs'
+import { readFile } from 'node:fs/promises'
+import { RoundData } from '../lib/round.js'
+import { evaluate } from '../lib/evaluate.js'
+import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
+import { fetchRoundDetails } from '../lib/spark-api.js'
+import createDebug from 'debug'
+import { Point } from '@influxdata/influxdb-client'
+import { createHash } from 'node:crypto'
+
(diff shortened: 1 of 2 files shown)


bajtos:
I'll review this PR next week. I am afraid there will be a merge conflict with https://github.com/filecoin-station/spark-evaluate/pull/396 😢 How should we proceed with this?

bin/evaluate-measurements.js
```diff
@@ -0,0 +1,176 @@
...
+const measurementsFile = await readFile(measurementsPath, 'utf8')
+for (const line of measurementsFile.split('\n').filter(Boolean)) {
+  const { roundIndex: _roundIndex, measurement } = JSON.parse(line)
+  const roundIndex = BigInt(_roundIndex)
+  if (!rounds.has(roundIndex)) rounds.set(roundIndex, [])
+  rounds.get(roundIndex).push(measurement)
+}
+
+const EVALUATION_FILE = `evaluation-${
+  createHash('sha256').update(measurementsFile).digest('hex').substring(0, 7)
```
bajtos, line 40: Working with filenames with a (random) hash can be cumbersome, e.g. when inspecting multiple miners and relying on shell auto-completion.

Have you considered building the output filename from the input filename by adding a suffix?

E.g. `measurements-f0aaa.ndjson` -> `measurements-f0aaa-evaluated.txt`

bajtos:
Before your change, the default behaviour is to not show rejected measurements in the output file. IIUC the proposed version, the first step will produce an intermediate file that includes rejected measurements too. I think this can create confusion when people not familiar with the concept of accepted/rejected scores use these tools.

An idea to consider:
- The first step creates a file with a suffix like `-raw` or `-all`
- The second step either creates files with no suffix (i.e. preserve the current output filenames) or use a different suffix like `-accepted` or `-evaluated`.


WDYT?

bin/evaluate-measurements.js
juliangruber: great idea!

https://github.com/filecoin-station/spark-evaluate/pull/401/commits/e2b5663f882b2bac2e74196540ce3bd028f07903

juliangruber:
> Before your change, the default behaviour is to not show rejected measurements in the output file. IIUC the proposed version, the first step will produce an intermediate file that includes rejected measurements too. I think this can create confusion when people not familiar with the concept of accepted/rejected scores use these tools.
> 
> An idea to consider:
> 
> * The first step creates a file with a suffix like `-raw` or `-all`
> * The second step either creates files with no suffix (i.e. preserve the current output filenames) or use a different suffix like `-accepted` or `-evaluated`.
> 
> WDYT?

Since https://github.com/filecoin-station/spark-evaluate/pull/401/commits/e2b5663f882b2bac2e74196540ce3bd028f07903, the output file will have `.evaluation.txt` suffix, which I think is great. I don't think there's any expectation about the shape of the intermediary file, it is supposed to contain all measurements submitted (in my mental model) and that it does.

juliangruber:
> I'll review this PR next week. I am afraid there will be a merge conflict with #396 😢
> 
> How should we proceed to handle that in the least painful way?

The merge conflict isn't going to be hard to deal with if we reapply https://github.com/filecoin-station/spark-evaluate/pull/396/files on top of this PR. Therefore I propose to merge this one first.

bajtos:
> > I'll review this PR next week. I am afraid there will be a merge conflict with #396 😢
> > How should we proceed to handle that in the least painful way?
> 
> The merge conflict isn't going to be hard to deal with if we reapply https://github.com/filecoin-station/spark-evaluate/pull/396/files on top of this PR. Therefore I propose to merge this one first.

I was thinking about this some more, and would like to rework #396 - see https://github.com/filecoin-station/spark-evaluate/pull/396#discussion_r1841572242

I agree to land this pull request first.

bajtos (approved):
I quickly skimmed through the changes and I don't see any obvious problems. Let's land this and then fix any issues later as we discover them.

:shipit:

bajtos:
> > Before your change, the default behaviour is to not show rejected measurements in the output file. IIUC the proposed version, the first step will produce an intermediate file that includes rejected measurements too. I think this can create confusion when people not familiar with the concept of accepted/rejected scores use these tools.
> > An idea to consider:
> > 
> > * The first step creates a file with a suffix like `-raw` or `-all`
> > * The second step either creates files with no suffix (i.e. preserve the current output filenames) or use a different suffix like `-accepted` or `-evaluated`.
> > 
> > WDYT?
> 
> Since [e2b5663](https://github.com/filecoin-station/spark-evaluate/commit/e2b5663f882b2bac2e74196540ce3bd028f07903), the output file will have `.evaluation.txt` suffix, which I think is great. I don't think there's any expectation about the shape of the intermediary file, it is supposed to contain all measurements submitted (in my mental model) and that it does.

I believe the old version included evaluation results in the measurements written to the per-miner file. After your change, the first script that fetches measurements produces a file that does not have information about the evaluation result. The second script produces only TXT file, so there is no way how to further analyse the processed per-miner measurements using `jq` or other tools. 

Please correct me if I'm wrong.

I think this can be also iterated on in a follow-up pull request.

Please remember to update our docs (https://docs.filspark.com/troubleshooting-miner-score, the source is in Notion) after you land this change.

juliangruber:
> I believe the old version included evaluation results in the measurements written to the per-miner file. After your change, the first script that fetches measurements produces a file that does not have information about the evaluation result. The second script produces only TXT file, so there is no way how to further analyse the processed per-miner measurements using `jq` or other tools.
> 
> Please correct me if I'm wrong.

Now, the measurements fetching prints a summary to stdout and produces a `.ndjson` file of raw measurements.

The evaluation script prints a summary to stdout and produces a `.txt` file of per-measurement evaluation. I will add a `.ndjson` file so that this data is machine processable.

PR #402, opened 2024-11-12
bin: add `fetch-historic-measurements`

juliangruber:
Based on https://github.com/filecoin-station/spark-evaluate/pull/401

### DIFF
--- a/bin/fetch-historic-measurements.js
+++ b/bin/fetch-historic-measurements.js
@@ -174,7 +174,7 @@ for (const [r, c] of Object.entries(resultCounts)) {
 }
 
 if (allMeasurementsWriter) {
-console.error('Wrote (ALL) raw measurements to %s', ALL_MEASUREMENTS_FILE)
+  console.error('Wrote (ALL) raw measurements to %s', ALL_MEASUREMENTS_FILE)
 }
 if (minerDataWriter) {
   console.error('Wrote (minerId=%s) raw measurements to %s', minerId, MINER_DATA_FILE)
--- a/bin/fetch-historic-measurements.js
+++ b/bin/fetch-historic-measurements.js
@@ -1,0 +1,289 @@
+/*
+Usage:
+
+1. Setup port forwarding between your local computer and Postgres instance hosted by Fly.io
+  ([docs](https://fly.io/docs/postgres/connecting/connecting-with-flyctl/)). Remember to use a
+  different port if you have a local Postgres server for development!
+   ```sh
+   fly proxy 5454:5432 -a spark-db
+   ```
+
+2. Find spark-db entry in 1Password and get the user and password from the connection string.
+
+3. Run the following command to fetch all measurements, remember to replace "user" and "password"
+   with the real credentials:
+
+   ```sh
+   DATABASE_URL="postgres://user:password@localhost:5454/spark" node bin/fetch-historic-measurements.js <range-start> <range-end>
+   ```
+
+   This will fetch all measurements committed between range-start (inclusive) and range-end (exclusive)
+   and write them to file.
+*/
+
+// dotenv must be imported before importing anything else
+import 'dotenv/config'
+
+import { Point } from '@influxdata/influxdb-client'
+import * as Sentry from '@sentry/node'
+import createDebug from 'debug'
+import fs from 'node:fs'
+import { mkdir, readFile, writeFile } from 'node:fs/promises'
+import os from 'node:os'
+import path from 'node:path'
+import pMap from 'p-map'
+import { fetchMeasurements, preprocess } from '../lib/preprocess.js'
+import { RoundData } from '../lib/round.js'
+import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
+import pg from 'pg'
+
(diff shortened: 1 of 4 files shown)


bin/fetch-historic-measurements.js
```diff
@@ -0,0 +1,289 @@
...
+  if (r === 'total') continue
+  console.log('  %s %s (%s%)',
+    r.padEnd(40),
+    String(c).padEnd(10),
+    Math.floor(c / resultCounts.total * 10000) / 100
+  )
+}
+
+if (allMeasurementsWriter) {
+console.error('Wrote (ALL) raw measurements to %s', ALL_MEASUREMENTS_FILE)
```
juliangruber, line 177:
```suggestion
  console.error('Wrote (ALL) raw measurements to %s', ALL_MEASUREMENTS_FILE)
```

bin/fetch-historic-measurements.js
```diff
@@ -0,0 +1,289 @@
...
+import path from 'node:path'
+import pMap from 'p-map'
+import { fetchMeasurements, preprocess } from '../lib/preprocess.js'
+import { RoundData } from '../lib/round.js'
+import * as SparkImpactEvaluator from '@filecoin-station/spark-impact-evaluator'
+import pg from 'pg'
+
+const { DATABASE_URL } = process.env
+
+Sentry.init({
```
bajtos, line 42: Why does this script need to report errors to Sentry? Do you envision we will be running it automatically in an unobserved way? If not, can we remove Sentry integration?

bin/fetch-historic-measurements.js
```diff
@@ -0,0 +1,289 @@
...
+   fly proxy 5454:5432 -a spark-db
+   ```
+
+2. Find spark-db entry in 1Password and get the user and password from the connection string.
+
+3. Run the following command to fetch all measurements, remember to replace "user" and "password"
+   with the real credentials:
+
+   ```sh
+   DATABASE_URL="postgres://user:password@localhost:5454/spark" node bin/fetch-historic-measurements.js <range-start> <range-end>
```
bajtos, line 17:
```suggestion
   DATABASE_URL="postgres://user:password@localhost:5454/spark" node bin/fetch-historic-measurements.js [contract-address] <range-start> <range-end>
```

Based on the code below.

```js
if (args.length === 0 || !args[0].startsWith('0x')) {
  args.unshift(SparkImpactEvaluator.ADDRESS)
}
```

bin/fetch-historic-measurements.js
```diff
@@ -0,0 +1,289 @@
...
+
+const USAGE = `
+Usage:
+  ${nodePath} ${selfPath} [contract-address] range-start range-end [minerId]
+
+Example:
+  ${nodePath} ${selfPath} 2024-10-01T00:00:00Z 2024-10-02T00:00:00Z
+`
+
+if (!contractAddress) {
```
bajtos, line 68: I think this will never happen because

```js
args.unshift(SparkImpactEvaluator.ADDRESS)
```

bin/fetch-historic-measurements.js
```diff
@@ -0,0 +1,289 @@
...
+try {
+  for (const { roundIndex, measurementCids } of rounds) {
+    signal.throwIfAborted()
+    await processRound(roundIndex, measurementCids, resultCounts)
+  }
+} catch (err) {
+  if (!signal.aborted) {
+    throw err
+  }
+}
```
bajtos, line 160: IIRC, we wanted to decouple "fetch measurements" from "evaluate fetched measurements".

What do you think about modifying this script to produce output suitable for consumption by `bin/evaluate-measurements` (introduced by #401)?

bajtos (changes requested):

bin/fetch-historic-measurements.js
juliangruber: 25d0286

bin/fetch-historic-measurements.js
bajtos: Oh, so the "fetch" part is responsible for preprocessing? I was expecting the fetch part to only fetch the measurements. I need to think about this a bit more.

bin/fetch-historic-measurements.js
```diff
@@ -14,7 +14,7 @@ Usage:
    with the real credentials:
 
    ```sh
-   DATABASE_URL="postgres://user:password@localhost:5454/spark" node bin/fetch-historic-measurements.js <range-start> <range-end>
+   DATABASE_URL="postgres://user:password@localhost:5454/spark" [contract-address] node bin/fetch-historic-measurements.js <range-start> <range-end>
```
bajtos, line 17:
```suggestion
   DATABASE_URL="postgres://user:password@localhost:5454/spark" node bin/fetch-historic-measurements.js [contract-address] <range-start> <range-end>
```

bajtos (approved):
Almost there! Please fix the example invocation per my suggestion below. The rest LGTM, no further review is necessary.

bin/fetch-historic-measurements.js
juliangruber: Will address this comment next week

PR #434, opened 2024-12-23
Spark TTFB poc

pyropy:
- Adds retrieval time stats to public stats

Relates to https://github.com/space-meridian/roadmap/issues/208


juliangruber:
Quoting @pyropy:

> I left them (the PRs) as draft as I am unsure if there is a way to reference spark-evaluate commit from spark-stats and I am not sure if we should have another section for ttfb on dashboard

lib/committee.js
```diff
@@ -48,6 +48,7 @@ export class Committee {
   addMeasurement (m) {
     assert.strictEqual(m.cid, this.retrievalTask.cid, 'cid must match')
     assert.strictEqual(m.minerId, this.retrievalTask.minerId, 'minerId must match')
+    assert.strictEqual(m.roundId, this.retrievalTask.roundId, 'roundId must match')
```
juliangruber, line 51: why was this added?

lib/preprocess.js
```diff
@@ -11,10 +11,11 @@ const debug = createDebug('spark:preprocess')
 
 export class Measurement {
   /**
+   * @param {Partial<import('./round.js').RoundData>} r
```
juliangruber, line 14: Since we're already passing `round.pointerize` as the last argument, I suggest we also pass `round.index` instead of `round`, for consistency and to prevent redundancy

lib/public-stats.js
```diff
@@ -225,3 +226,37 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
     flatStats.map(stat => stat.retrievable)
   ])
 }
+
+/**
+ * @param {pg.Client} pgClient
+ * @param {Iterable<Committee>} committees
+ */
+const updateRetreivalTimes = async (pgClient, committees) => {
```
juliangruber, line 234:
```suggestion
const updateRetreivalTimings = async (pgClient, committees) => {
```

lib/public-stats.js
```diff
@@ -225,3 +226,37 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
...
+      ttfbMeasurments.push(m.first_byte_at - m.start_at)
+    }
+
+    const timeToFirstByteP50 = getValueAtPercentile(ttfbMeasurments, 0.5)
+    stats.push({ minerId, taskId, timeToFirstByteP50 })
+  }
+
+  // conflic should never happen, but in case it does we'll ignore the new value
+  await pgClient.query(`
+    INSERT INTO retrieval_times
```
juliangruber, line 253:
```suggestion
    INSERT INTO retrieval_timings
```

lib/public-stats.js
```diff
@@ -225,3 +226,37 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
...
+    for (const m of c.measurements) {
+      if (m.fraudAssessment !== 'OK') continue
+      ttfbMeasurments.push(m.first_byte_at - m.start_at)
+    }
+
+    const timeToFirstByteP50 = getValueAtPercentile(ttfbMeasurments, 0.5)
+    stats.push({ minerId, taskId, timeToFirstByteP50 })
+  }
+
+  // conflic should never happen, but in case it does we'll ignore the new value
```
juliangruber, line 251: If conflicts shouldn't happen, I suggest we remove the conflict handling and let it fail. If everything is right, it will never fail. If it fails, it will inform us of a bug to fix.

juliangruber (changes requested):

juliangruber:
> Quoting @pyropy:
> 
> > I left them (the PRs) as draft as I am unsure if there is a way to reference spark-evaluate commit from spark-stats and I am not sure if we should have another section for ttfb on dashboard

I'm marking this as ready for review as the open questions don't affect this PR and I think it can be merged as soon as reviews pass

lib/committee.js
pyropy: I have changed task id structure from `cid::minerId` to `cid::minerId::roundId` hence we're checking round id here.

juliangruber: I understand, but what is the motivation?

lib/public-stats.js
```diff
@@ -225,3 +226,37 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
     flatStats.map(stat => stat.retrievable)
   ])
 }
+
+/**
+ * @param {pg.Client} pgClient
+ * @param {Iterable<Committee>} committees
+ */
+const updateRetreivalTimings = async (pgClient, committees) => {
```
bajtos, line 234: Typo.

```suggestion
const updateRetrievalTimings = async (pgClient, committees) => {
```

lib/public-stats.js
```diff
@@ -225,3 +226,37 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
...
+    const timeToFirstByteP50 = getValueAtPercentile(ttfbMeasurments, 0.5)
+    stats.push({ minerId, taskId, timeToFirstByteP50 })
+  }
+
+  // conflic should never happen, but in case it does we'll ignore the new value
+  await pgClient.query(`
+    INSERT INTO retrieval_timings
+    (day, miner_id, task_id, time_to_first_byte_p50) VALUES 
+    (now(), unnest($1::text[]), unnest($2::text[]), unnest($3::int[]))
+    ON CONFLICT(day, miner_id, task_id) DO NOTHING
```
bajtos, line 256: I have mixed feelings about this design.

- Task id was designed to distinguish tasks within one round. spark-evaluate always looks at one round only.
- Since it's only `updateRetreivalTimings()` that needs to handle the case when one task is performed more than once during a day, I prefer to implement a solution that's limited to `updateRetreivalTimings()` only. For example, we can forward the current round number through `updatePublicStats()` to `updateRetreivalTimings()` and then combine old-style taskId with the round number.
- Because task_id includes a round number, it does not help us detect cases when the same content was tested twice on the same day, it only ensures we can record timings for each task occurrence.

I propose a different DB schema for consideration: Instead of having one row per day+task+round, have only one row per day and store the p50 values in an array.

Something along the following lines:

```sql
INSERT INTO retrieval_timings
(day, miner_id, time_to_first_byte_p50) VALUES 
(now(), unnest($2::text[]), unnest($3::int[]))
ON CONFLICT(day, miner_id) 
DO UPDATE SET
  time_to_first_byte_p50 = array_cat(
    retrieval_timings.time_to_first_byte_p50,
    EXCLUDED.time_to_first_byte_p50
  )
```

migrations/021.do.add-retrieval-times.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE retrieval_times (
+  day DATE NOT NULL,
+  miner_id TEXT NOT NULL,
+  task_id TEXT NOT NULL,
+  time_to_first_byte_p50 INT NOT NULL,
```
bajtos, line 5: This column name is rather long, how about using the abbreviation TTFB?

```suggestion
  ttfb_p50 INT NOT NULL,
```

test/public-stats.test.js
```diff
@@ -534,8 +535,88 @@ describe('public-stats', () => {
...
     return today
   }
+
+  /**
+   *
+   * @param {Measurement} measurment
+   * @param {number} timeToFirstByte  Time in milliseconds
+   * @returns
+   */
+  function givenTimeToFirstByte (measurment, timeToFirstByte) {
```
bajtos, line 618: Typo

```suggestion
  function givenTimeToFirstByte (measurement, timeToFirstByte) {
```

bajtos (changes requested):

lib/public-stats.js
pyropy: Nice idea, let's go with the array for the value field.

lib/committee.js
pyropy: Motivation behind adding the `roundId` to `taskId` was to avoid conflicts when updating `retrieval_timings` values as pkey for the value was day, taskId and minerId. If we'd kept old `taskId` in that structure we would encounter conflicts whenever we'd measure same CID for same storage provider twice in a day. 

I am going to switch table to use int array for storing p50 values for all measurements for a single storage provider and rollback changes I've done around new `taskId`.

lib/public-stats.js
```diff
@@ -227,3 +228,44 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
...
+    INSERT INTO retrieval_timings (day, miner_id, ttfb_p50) 
+    SELECT now(), miner_id, ttfb_p50 FROM jsonb_to_recordset($1::jsonb) AS t (miner_id text, ttfb_p50 int[])
+    ON CONFLICT (day, miner_id)
+    DO UPDATE SET
+      ttfb_p50 = array_cat(
+        retrieval_timings.ttfb_p50,
+        EXCLUDED.ttfb_p50
+      )
+  `, [
+    JSON.stringify(rows)
```
pyropy, line 269: I have done this as I am unable to `unnest` nested array (`unnest` would convert nested array to int values). If you have some suggestions on how to do this in a cleaner way I'd appreciate that 🙏🏻

bajtos (approved):
Almost there! 💪🏻

bajtos (changes requested):
Ouch, I meant to request changes.

lib/public-stats.js
```diff
@@ -227,3 +228,44 @@ const updateDailyDealsStats = async (pgClient, committees, findDealClients) => {
...
+const updateRetrievalTimings = async (pgClient, committees) => {
+  /** @type {Map<string, number[]>} */
+  const retrievalTimings = new Map()
+  for (const c of committees) {
+    if (!c.evaluation || !c.evaluation.hasRetrievalMajority || c.evaluation.retrievalResult !== 'OK') continue
+    const { minerId } = c.retrievalTask
+    const ttfbMeasurments = []
+    for (const m of c.measurements) {
+      if (m.fraudAssessment !== 'OK' || !m.first_byte_at || !m.start_at) continue
+      ttfbMeasurments.push(m.first_byte_at - m.start_at)
```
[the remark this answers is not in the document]
pyropy: Nice catch, I'm going to add test for this also.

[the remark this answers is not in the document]
bajtos: Please also add a check that `m.first_byte_at - m.start_at` is not `NaN`. That should hopefully cover the remaining edge cases.

```
❯ node
Welcome to Node.js v22.13.0.
Type ".help" for more information.
> 'a' - 12
NaN
```

lib/public-stats.js
pyropy: I've tried removing `JSON.stringify` but it throws following error `error: invalid input syntax for type json`. I think node-postgres handles serialization but only to postgres binary format, so we need to serialize it to json first before it get's converted to binary.

lib/public-stats.js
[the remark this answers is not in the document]
pyropy: I have added checks and test to cover edge cases.

test/public-stats.test.js
```diff
@@ -564,13 +564,15 @@ describe('public-stats', () => {
...
         givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1first', retrievalResult: 'OK' }, 1000),
-        givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1first', retrievalResult: 'OK' }, 2000),
         givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1first', retrievalResult: 'OK' }, 3000),
         givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1second', retrievalResult: 'OK' }, 2000),
         givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1second', retrievalResult: 'OK' }, 1000),
-        givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 
+        // measurments with invalid values
+        givenTimeToFirstByte({ ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1second', retrievalResult: 'OK' }, -1000),
+        // @ts-ignore
+        { ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1second', retrievalResult: 'OK', first_byte_at: 'invalid' }
```
bajtos, line 575: Nitpick - this may be safer as it disables the type check only for one property value, not the entire line.

```suggestion
        // @ts-ignore
        { ...VALID_MEASUREMENT, cid: 'cidone', minerId: 'f1second', retrievalResult: 'OK', first_byte_at: /** @type {any} */('invalid') }
```

bajtos (approved):
LGTM 👏🏻

PR #437, opened 2025-01-07
feat: metric for successful http retrievals

NikolasHaimerl:
This MR proposes the following changes: 

1. Add a metric for counting the successful http retrievals

Following the definition of a this [issue](https://github.com/space-meridian/roadmap/issues/192#issue-2668841168)  the metric is defined as:

> how many retrievals succeeded using HTTP protocol for CAR transfer. (A retrieval that had to use Graphsync because the SP does not advertise HTTP is considered as a failure.)

In practice a successful http request is indicated by `Measurement.retrievalResult`  as well as the protocol being used indicated by`Measurement.protocol`.

### DIFF
--- a/test/retrieval-stats.test.js
+++ b/test/retrieval-stats.test.js
@@ -19,7 +19,8 @@ describe('retrieval statistics', () => {
     /** @type {Measurement[]} */
     const measurements = [
       {
-        ...VALID_MEASUREMENT
+        ...VALID_MEASUREMENT,
+        protocol: 'http'
       },
       {
         ...VALID_MEASUREMENT,
@@ -31,7 +32,8 @@ describe('retrieval statistics', () => {
         first_byte_at: new Date('2023-11-01T09:00:10.000Z').getTime(),
         end_at: new Date('2023-11-01T09:00:50.000Z').getTime(),
         finished_at: new Date('2023-11-01T09:00:30.000Z').getTime(),
-        byte_length: 2048
+        byte_length: 2048,
+        protocol: 'http'
       },
       {
         ...VALID_MEASUREMENT,
@@ -57,7 +59,7 @@ describe('retrieval statistics', () => {
         // invalid task
         cid: 'bafyinvalid',
         provider_address: '/dns4/production-ipfs-peer.pinata.cloud/tcp/3000/ws/p2p/Qma8ddFEQWEU8ijWvdxXm3nxU7oHsRtCykAaVz8WUYhiKn',
-        protocol: 'bitswap'
+        protocol: 'http'
       }
     ]
 
@@ -68,6 +70,8 @@ describe('retrieval statistics', () => {
     assertPointFieldValue(point, 'measurements', '4i')
     assertPointFieldValue(point, 'unique_tasks', '3i')
     assertPointFieldValue(point, 'success_rate', '0.25')
+    assertPointFieldValue(point, 'success_rate_http', '0.25')
+    assertPointFieldValue(point, 'participants', '2i')
     assertPointFieldValue(point, 'participants', '2i')
     assertPointFieldValue(point, 'inet_groups', '2i')
     assertPointFieldValue(point, 'download_bandwidth', '209718272i')
--- a/lib/retrieval-stats.js
+++ b/lib/retrieval-stats.js
@@ -58,6 +58,7 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
   const ttfbValues = []
   const durationValues = []
   const sizeValues = []
+  let httpSuccesses = 0
 
   for (const m of measurements) {
     // `retrievalResult` should be always set by lib/preprocess.js, so we should never encounter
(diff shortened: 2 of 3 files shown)


test/retrieval-stats.test.js
```diff
@@ -57,7 +59,8 @@ describe('retrieval statistics', () => {
         // invalid task
         cid: 'bafyinvalid',
         provider_address: '/dns4/production-ipfs-peer.pinata.cloud/tcp/3000/ws/p2p/Qma8ddFEQWEU8ijWvdxXm3nxU7oHsRtCykAaVz8WUYhiKn',
-        protocol: 'bitswap'
+        // Check that the task is ignored if the retrieval result is not OK
+        protocol: 'http'
```
bajtos, line 63: Have you seen this test fail?

My understanding is that this measurement is ignored because it's for an invalid task - a task that's not a part of this round.

We already have a different measurement in the test data that's for `http` protocol and reports a failure - see the second item with `retrievalResult: 'TIMEOUT'`.

test/retrieval-stats.test.js
```diff
@@ -31,7 +32,8 @@ describe('retrieval statistics', () => {
         first_byte_at: new Date('2023-11-01T09:00:10.000Z').getTime(),
         end_at: new Date('2023-11-01T09:00:50.000Z').getTime(),
         finished_at: new Date('2023-11-01T09:00:30.000Z').getTime(),
-        byte_length: 2048
+        byte_length: 2048,
+        protocol: 'http'
```
bajtos, line 36: At the top of this measurement, we set `indexerResult: 'HTTP_NOT_ADVERTISED'`.

The combination of `indexerResult: 'HTTP_NOT_ADVERTISED'` and `protocol: 'http'` cannot happen in practice.

This test is verifying `indexer_rate_HTTP_NOT_ADVERTISED` and `rate_of_deals_advertising_http`, so we need to keep a measurement with `indexerResult: 'HTTP_NOT_ADVERTISED'`.

I propose the following: Revert the changes in this test and add a new test focused solely on `success_rate_http`.

In my experience, tests focused on a smaller scope are easier to maintain. Tests covering too much (like this one) are difficult to modify, as you can see yourself. I created this test as a quick solution to get basic test coverage - at that time, it was a good tradeoff. Now it's time to stop expanding this test and test new features in a cleaner way.

bajtos (changes requested):
Great start! Let's improve the tests a bit.

test/retrieval-stats.test.js
NikolasHaimerl: Alright, I'll create a dedicated test for it. Since most values of the measurments are strings and not enums I didn't want to make any assumptions on what combination can or cannot happen and solely go off the logic that goes into calculating a successful http response.

test/retrieval-stats.test.js
NikolasHaimerl: I removed this part in favour of a dedicated test for successful http requests.

bajtos (changes requested):
Almost there!

test/retrieval-stats.test.js
```diff
@@ -213,6 +209,39 @@ describe('retrieval statistics', () => {
...
+        ...VALID_MEASUREMENT
+      },
+      {
+        // A successful http measurement
+        ...VALID_MEASUREMENT,
+        protocol: 'http'
+      },
+      {
+        ...VALID_MEASUREMENT,
+        retrievalResult: 'HTTP_500'
```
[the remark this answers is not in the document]
NikolasHaimerl: I want to test that if the protocol is http but the retrievalResult is not `OK` then the measurment should not count as a success.

test/retrieval-stats.test.js
```diff
@@ -208,6 +208,41 @@ describe('retrieval statistics', () => {
...
+        ...VALID_MEASUREMENT,
+        protocol: 'http',
+        retrievalResult: 'HTTP_500'
+      },
+      {
+        ...VALID_MEASUREMENT,
+        // Should not be picked up, as the retrieval timed out
+        retrievalResult: 'TIMEOUT',
+        protocol: 'http'
+      }
```
bajtos, line 235: These two items seem to test the same test case, but that's a detail.

bajtos (approved):
LGTM! 👏🏻

bajtos:
@NikolasHaimerl Please go ahead and land the pull request yourself.

Our GitHub Actions CI/CD pipeline will automatically deploy this change to Fly.io. After the current Spark round ends and spark-evaluate finishes the evaluation run, you will see the first new data point in InfluxDB.

test/retrieval-stats.test.js
NikolasHaimerl: I wanted to make sure that the result error is independent of the result. As soon as there is an error -> the http retrieval is marked as unsuccessful.

juliangruber (commented):
Great!

PR #439, opened 2025-01-08
feat: assert indexer_result during preprocessing

bajtos:
Move the validation step earlier to the preprocess/evaluate pipeline.  Measurements missing the indexer result are dropped as invalid.

This is the first step to refactor `fraudAssessment` field for more clarity, see https://github.com/filecoin-station/spark-evaluate/pull/396#discussion_r1850597708

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -268,15 +268,6 @@ export const runFraudDetection = async ({
   )
 
   //
-  // 0. Filter out measurements missing required fields.
-  //
-  for (const m of measurements) {
-    if (!m.indexerResult) {
-      m.fraudAssessment = 'IPNI_NOT_QUERIED'
-    }
-  }
-
-  //
   // 1. Filter out measurements not belonging to any valid task in this round
   //    or missing some of the required fields like `inet_group`
   //
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -230,7 +230,7 @@ export const parseParticipantAddress = filWalletAddress => {
 /**
  * @param {Measurement} measurement
  */
-const assertValidMeasurement = measurement => {
+export const assertValidMeasurement = measurement => {
   assert(
     typeof measurement === 'object' && measurement !== null,
     'object required'
@@ -238,6 +238,7 @@ const assertValidMeasurement = measurement => {
   assert(ethers.isAddress(measurement.participantAddress), 'valid participant address required')
   assert(typeof measurement.inet_group === 'string', 'valid inet group required')
   assert(typeof measurement.finished_at === 'number', 'field `finished_at` must be set to a number')
+  assert(measurement.indexerResult, 'field `indexerResult` must be set')
   if (measurement.stationId) {
     assert(
       typeof measurement.stationId === 'string' &&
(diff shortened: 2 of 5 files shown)


lib/evaluate.js
```diff
@@ -267,15 +267,6 @@ export const runFraudDetection = async ({
...
-  // 0. Filter out measurements missing required fields.
-  //
-  for (const m of measurements) {
-    if (!m.indexerResult) {
-      m.fraudAssessment = 'IPNI_NOT_QUERIED'
-    }
-  }
-
   //
   // 1. Filter out measurements not belonging to any valid task in this round
```
NikolasHaimerl, line 271: Do we want to start with index 1. now?

lib/preprocess.js
```diff
@@ -230,14 +230,15 @@ export const parseParticipantAddress = filWalletAddress => {
...
-const assertValidMeasurement = measurement => {
+export const assertValidMeasurement = measurement => {
   assert(
     typeof measurement === 'object' && measurement !== null,
     'object required'
   )
   assert(ethers.isAddress(measurement.participantAddress), 'valid participant address required')
   assert(typeof measurement.inet_group === 'string', 'valid inet group required')
   assert(typeof measurement.finished_at === 'number', 'field `finished_at` must be set to a number')
+  assert(measurement.indexerResult, 'field `indexerResult` must be set')
```
NikolasHaimerl, line 241: How come that before this change measurments could have a missing `indexerResult` but after this PR we expect it to always be set?

bajtos: Great question!

Initially, there was no `indexerResult` field reported by Spark checker nodes.

Then we added `indexerResult` everywhere, but some nodes were not reporting that field for a while - I suspect somebody was running a modified version of the checker node.

At that point, I introduced the fraud-assessment step to flag measurements missing `indexerResult` field. This way, it was more difficult for operators of the misbehaving nodes to notice the change - their measurements were accepted but not rewarded. 

Eventually, they fixed their nodes. 

Nowadays, all measurements include `indexerResult`. You can check this yourself in [Spark Internal Dashboard](https://spacemeridian.grafana.net/d/c4bd3c80-9360-4ea3-8b75-c5ecea907659/spark-internal-dashboard) by editing the panel `Retrieval Status (Honest)` and changing the queried measurement name from `retrieval_stats_honest` to `retrieval_stats_all`.
Screenshot 2025-01-08 at 14 07 28

We could change spark-api to enforce this constraint and reject requests submitting measurements with missing `indexerResult`, but I think that's not necessary.

https://github.com/filecoin-station/spark-api/blob/65616e7d7b2bcf29ece042bbe23e221d17aaf794/api/index.js#L82

bajtos: In other words, measurements already contain the `indexerResult` field, irrespective of this pull request.

What this pull request changes is how we handle the unlikely (but theoretically possible) case when a measurement does not have `indexerResult`.

- Before: we flagged the measurement during fraud detection, and we had visibility into the number of measurements rejected this way
- After: we reject such measurements in the preprocessing step, simplifying the rest of the evaluation workflow.

NikolasHaimerl: I see, thanks for explaining.

NikolasHaimerl (approved):

juliangruber (approved):
great!

PR #440, opened 2025-01-08
feat: Add HTTP-RSR to daily retrieval stats

NikolasHaimerl:
This MR proposes the following changes:

1. Add successful http retrieval counter to `retrieval_stats` table
2. Keep track of successful http retrievals on a daily timeframe
3. Tests for processing the success of daily http retrievals


lib/provider-retrieval-result-stats.js
```diff
@@ -16,15 +16,20 @@ const withPgClient = fn => async ({ createPgClient, ...args }) => {
...
     // IMPORTANT: include minority results in the calculation
     for (const m of c.measurements) {
       const minerId = m.minerId
-      const retrievalStats = providerRetrievalResultStats.get(minerId) ?? { total: 0, successful: 0 }
+      const retrievalStats = providerRetrievalResultStats.get(minerId) ?? { total: 0, successful: 0, successfulHttp: 0 }
       retrievalStats.total++
-      if (m.retrievalResult === 'OK') retrievalStats.successful++
+      if (m.retrievalResult === 'OK') {
+        retrievalStats.successful++
+        if (m.protocol && m.protocol.toLowerCase() === 'http') {
```
bajtos, line 29: Is worth moving the normalization step (`.toLowerCase()`) to a single place to avoid duplication?

https://github.com/filecoin-station/spark-evaluate/blob/679265d23b754e9c6215f7c91c7bce4dc8626426/lib/preprocess.js#L30

If you agree, then make sure the normalife the value before calling `pointerize`.

```js
this.protocol = pointerize(m.protocol?.toLowerCase())
```

You can learn more about `pointerize` here: https://blog.filstation.app/posts/how-we-reduced-memory-usage-by-90#block-b612514f9909412190bfefe23b03826a

migrations/001.do.retrieval-stats.sql
```diff
@@ -1,5 +1,6 @@
 CREATE TABLE retrieval_stats (
   day DATE NOT NULL PRIMARY KEY,
   total INT NOT NULL,
-  successful INT NOT NULL
+  successful INT NOT NULL,
+  successful_http INT NOT NULL
 );
```
bajtos, line 6: DB migration scripts are immutable. You need to add a new SQL script to alter the table and add the new column.

Sadly, I could not find a good resource explaining this concept. 

We use https://www.npmjs.com/package/postgrator to run the migration scripts for us. 

[Evolutionary Database Design](https://martinfowler.com/articles/evodb.html) explains the practice in wider context.

The key insight is that we already have `retrieval_stats` table in the production DB. When deploying this pull request, we need to modify the schema to add the new column, preserving existing data (historical retrieval stats).

Your task is to write such script. Postgrator will take care of the rest.

test/evaluate.js
```diff
@@ -96,10 +96,11 @@ describe('evaluate', async function () {
       day: today(),
       miner_id: VALID_TASK.minerId,
       total: 1,
-      successful: 1
+      successful: 1,
+      // None of the measurments use http
```
bajtos, line 100: Typo?

```suggestion
      // None of the measurements use http
```

bajtos (changes requested):
Great start!

lib/provider-retrieval-result-stats.js
```diff
@@ -26,9 +26,7 @@ export const build = committees => {
       retrievalStats.total++
       if (m.retrievalResult === 'OK') {
         retrievalStats.successful++
-        if (m.protocol && m.protocol.toLowerCase() === 'http') {
-          retrievalStats.successfulHttp++
-        }
+        if (m.protocol && m.protocol === 'http') { retrievalStats.successfulHttp++ }
```
bajtos, line 29: 👏🏻

migrations/001.do.retrieval-stats.sql
NikolasHaimerl: Thanks for the tip, I included an appropriate migration script.

lib/provider-retrieval-result-stats.js
NikolasHaimerl: Great piece on optimizing memory consumption. I made the changes accordingly.

migrations/021.do.sucessful-http-retrieval-stats.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE retrieval_stats
+ADD COLUMN IF NOT EXISTS successful_http INT NOT NULL DEFAULT 0
```
bajtos, line 2: Nitpick: missing EOL at EOF.

bajtos, line 2: I am wondering if setting `successful_http = 0` is the right approach for historical data. The real RSR was almost certainly non-zero, probably in the range of ~5%.

How about making this column nullable?

```suggestion
ADD COLUMN IF NOT EXISTS successful_http INT
```

This way, spark-stats REST API handler can exclude HTTP-RSR for days where we did not capture it. As a result, historical charts will not show any HTTP-RSR value for those days. (In your current proposal, historical charts would show 0%.)

bajtos (commented):
Almost there!

migrations/021.do.sucessful-http-retrieval-stats.sql
NikolasHaimerl: Good idea, I made the change.

bajtos (approved):
:shipit:

migrations/001.do.retrieval-stats.sql
juliangruber: See https://en.wikipedia.org/wiki/Schema_migration

juliangruber: I'm aware of this concept from Rails https://guides.rubyonrails.org/active_record_migrations.html

PR #442, opened 2025-01-09
refactor: break fraudAssessment into evaluations

bajtos:
Break `Measurement.fraudAssessment` into two new fields:

- `taskingEvaluation` for the tasking-related checks
  Example: DUP_INET_GROUP

- `consensusEvaluation` for the result of the majority seeking process based on committees.
  Example: MINORITY_RESULT

After this change, places filtering "accepted" measurements have to explicitly spell out how they define "accepted".

- Some places are interested in tasking evaluation results only and consider minority results as "accepted" too. Example: RSR calculated from individual measurements.

- Other places are stricter and want only measurements in majority.  Example: which measurements to reward.

However, this pull request is intended to be pure refactoring with no change in the functionality. It should simply expand the check `fraudAssessment === 'OK'` into `taskingEvaluation === 'OK' && consensusEvaluation === 'MAJORITY_RESULT'` and surface the places where we may want to include minority measurements too. Such changes can be implement in follow-up pull requests.

This is a follow-up for https://github.com/filecoin-station/spark-evaluate/pull/396#discussion_r1850597708

See also #439


lib/evaluate.js
```diff
@@ -62,12 +62,14 @@ export const evaluate = async ({
     requiredCommitteeSize,
     logger
   })
-  const honestMeasurements = measurements.filter(m => m.fraudAssessment === 'OK')
+  const measurementsToReward = measurements.filter(
+    m => m.taskingEvaluation === 'OK' && m.majorityEvaluation === 'OK'
+  )
```
bajtos, line 67: TODO: add a test to verify that we don't reward measurements that are not in majority.

lib/evaluate.js
```diff
@@ -152,7 +154,11 @@ export const evaluate = async ({
   try {
     recordTelemetry('retrieval_stats_honest', (point) => {
       point.intField('round_index', roundIndex)
-      buildRetrievalStats(honestMeasurements, point)
+      // DISCUSSION POINT: should we include minority measurements in these stats?
+      // Right now, we include majority measurements only. As a result,
+      // the retrieval-result-breakdown does not show outlier result types
+      // like infrequently-encountered HTTP error codes.
+      buildRetrievalStats(measurementsToReward, point)
```
bajtos, line 161: TODO: resolve this discussion.

> Should we include minority measurements in these stats?
> Right now, we include majority measurements only. As a result,
> the retrieval-result-breakdown does not show outlier result types
> like infrequently-encountered HTTP error codes.

I am leaning towards including minority results in "retrieval_stats_honest", but implementing such change in a new pull request.

bin/evaluate-measurements.js
```diff
@@ -98,22 +98,22 @@ async function processRound (roundIndex, measurements, resultCounts) {
...
+    if (m.taskingEvaluation !== 'OK') continue
     resultCounts.total++
     resultCounts[m.retrievalResult] = (resultCounts[m.retrievalResult] ?? 0) + 1
   }
 
   if (!keepRejected) {
     round.measurements = round.measurements
       // Keep accepted measurements only
-      .filter(m => m.fraudAssessment === 'OK')
+      .filter(m => m.taskingEvaluation === 'OK')
```
bajtos, line 109: This is fixing the bug I wanted to fix in https://github.com/filecoin-station/spark-evaluate/pull/396.

Are we okay with that, or should I preserve the current behaviour since refactorings are expected to not change observable behaviour?

lib/round.js
```diff
@@ -7,8 +9,9 @@ export class RoundData {
    */
   constructor (index) {
     this.index = index
-    /** @type {import('./preprocess.js').Measurement[]} */
+    /** @type {string[]} */
     this.measurementBatches = []
+    /** @type {Measurement[]} */
     this.measurements = []
```
bajtos, line 15: The field `measurementBatches` stores a list of CIDs, see here:

https://github.com/filecoin-station/spark-evaluate/blob/670984007ce2b64177281a1f98ebe611274c8132/lib/preprocess.js#L119

bajtos (commented):
🧵 TODO items blocking the conversion from "draft" to "ready" 👇🏻

bajtos (commented):
More TODOs 👇🏻

lib/evaluate.js
bajtos: see 1392633 and https://github.com/filecoin-station/spark-evaluate/issues/446

bin/evaluate-measurements.js
bajtos: See https://github.com/filecoin-station/spark-evaluate/issues/446

lib/evaluate.js
bajtos: Added in 11d896e.

Test failure when I comment out the majority-evaluation condition:

```
 1) evaluate
       rewards majority measurements only:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  [
+   '0x000000000000000000000000000000000000dEaD',
    '0x10',
    '0x20',
+   '0x30'
  ]

      + expected - actual

       [
      -  "0x000000000000000000000000000000000000dEaD"
         "0x10"
         "0x20"
      -  "0x30"
       ]
      
      at Context.<anonymous> (file:///Users/bajtos/src/spark/evaluate/test/evaluate.js:277:12)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

lib/platform-stats.js
```diff
@@ -67,7 +67,7 @@ export const updateStationsAndParticipants = async (
     }
 
     stationStats.total++
-    if (m.fraudAssessment === 'OK') stationStats.accepted++
+    if (m.taskingEvaluation === 'OK' && m.majorityEvaluation === 'OK') stationStats.accepted++
```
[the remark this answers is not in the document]
bajtos: Added in 01dc39f

Example test failure:

```
  1) platform-stats
       updateStationsAndParticipants
         counts only majority measurements as accepted:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  [
    {
+     accepted_measurement_count: 2,
-     accepted_measurement_count: 1,
      day: '2025-01-10',
      participant_id: 1,
      total_measurement_count: 2
    }
  ]

      + expected - actual

       [
         {
      -    "accepted_measurement_count": 2
      +    "accepted_measurement_count": 1
           "day": "2025-01-10"
           "participant_id": 1
           "total_measurement_count": 2
         }
```

lib/retrieval-stats.js
```diff
@@ -222,7 +223,7 @@ export const recordCommitteeSizes = (committees, point) => {
       data.nodes.add(`${m.inet_group}::${m.participantAddress}`)
       data.measurements++
 
-      if (m.fraudAssessment === 'OK') {
+      if (m.taskingEvaluation === 'OK' && m.majorityEvaluation === 'OK') {
```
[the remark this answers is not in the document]
bajtos: This is already covered by the following test:

https://github.com/filecoin-station/spark-evaluate/blob/c08c10aecf9fcc1f204b9792594a94e5d4fe9adc/test/retrieval-stats.test.js#L393-L433

NikolasHaimerl, line 226: Can't we also resuse `isAccepted` here?

bajtos: I thought about that, too. This code is in a different function, though. Also, the criteria defining which measurements are considered accepted can differ between these two places in the future, so I prefer not to reuse the condition.

bin/dry-run.js
```diff
@@ -148,7 +148,7 @@ if (ignoredErrors.length) {
 }
 
 if (DUMP) {
-  const props = ['cid', 'minerId', 'participantAddress', 'inet_group', 'retrievalResult', 'fraudAssessment']
+  const props = ['cid', 'minerId', 'participantAddress', 'inet_group', 'retrievalResult', 'taskingEvaluation']
```
juliangruber, line 151: should this not include `majorityEvaluation`, to represent the same range of information as the former `fraudAssessment`?
```suggestion
  const props = ['cid', 'minerId', 'participantAddress', 'inet_group', 'retrievalResult', 'taskingEvaluation', 'majorityEvaluation']
```

bin/evaluate-measurements.js
```diff
@@ -98,22 +98,24 @@ async function processRound (roundIndex, measurements, resultCounts) {
   })
 
   for (const m of round.measurements) {
-    if (m.fraudAssessment !== 'OK') continue
+    if (m.taskingEvaluation !== 'OK') continue
```
juliangruber, line 101: This is also a change in behavior. Previously, majority failures would be rejected here, now they are kept.

Before I continue with the review, can we please resolve this and the above comment? It looks like this question affects most of the changes.

bin/dry-run.js
juliangruber: I re-read the PR description and now see that this change in behavior might be intentional. Are you confident that all changes in behavior in this PR are intentional?

bajtos: Good catch, I'll accept your suggestion.

I'd like to point out that `dry-run` is a script we use internally only, and `DUMP` is an opt-in tool for debugging. I haven't used it for months. IMO, a breaking change here would have a very low impact and should be easy to fix.

bin/evaluate-measurements.js
bajtos: Great catch, I'll fix this place to preserve the current behaviour and add a FIXME comment similar to the one on lines 109-111 below.

I'd also like to point out that this is a helper script for SPs to inspect recent Spark measurements, it's known to be flawed (see #396), and yet nobody complained about that bug yet - so again, I think the impact of a breaking change in this file is low.

bajtos: >I re-read the PR description and now see that this change in behavior might be intentional. Are you confident that all changes in behavior in this PR are intentional?
> (...)
> Before I continue with the review, can we please resolve this and the above comment? It looks like this question affects most of the changes.

I started this PR with the intention to combine refactoring with changes fixing the conditions determining which measurements are considered as accepted. I quickly realise that would make it difficult to reason about the changes in the future, so I decided to pivot and change this pull request into pure refactoring with no change in functionality.

I updated the pull request description to clearly spell out this intention.

I am 90% confident that this pull request does not introduce any change in the functionality in the `lib` files. I wasn't paying as much attention to `bin` files, but I think you caught the remaining places to fix and this PR is not introducing any unintentional changes now.

bajtos: I searched for all occurences of `fraudAssessment ===` in the pull request and verified that they are all expanded to `taskingEvaluation === 'OK' && majorityEvaluation === 'OK'`.

I found one place I had to fix - see [16938d8](https://github.com/filecoin-station/spark-evaluate/pull/442/commits/16938d8df2565c0c0245a07dd3cebb39323c3667)

lib/evaluate.js
```diff
@@ -88,24 +90,24 @@ export const evaluate = async ({
...
     DUP_INET_GROUP: 0,
     TOO_MANY_TASKS: 0
   }
   for (const m of measurements) {
-    fraudAssessments[m.fraudAssessment] = (fraudAssessments[m.fraudAssessment] ?? 0) + 1
+    evaluationOutcomes[m.taskingEvaluation] = (evaluationOutcomes[m.taskingEvaluation] ?? 0) + 1
   }
   logger.log(
-    'EVALUATE ROUND %s: Evaluated %s measurements, found %s honest entries.\n%o',
+    'EVALUATE ROUND %s: Evaluated %s measurements, accepted %s entries to reward.\n%o',
```
juliangruber, line 106:
```suggestion
    'EVALUATE ROUND %s: Evaluated %s measurements, rewarding %s entries.\n%o',
```

To avoid confusion with "accepted measurements"

lib/evaluate.js
```diff
@@ -271,7 +275,7 @@ export const runFraudDetection = async ({
   //    or missing some of the required fields like `inet_group`
   //
   for (const m of measurements) {
-    if (m.fraudAssessment) continue
+    if (m.taskingEvaluation) continue
```
juliangruber, line 279: I believe in this first step no measurement can yet have a `takingEvaluation`

lib/platform-stats.js
juliangruber, line 70: We might want to refactor this and make it safer by adding a helper method `m#accepted()` or `m#ok()`. Wdyt?

lib/preprocess.js
```diff
@@ -22,8 +22,10 @@ export class Measurement {
     // Note: providerId is recorded by spark-publish but we don't use it for evaluations yet
     this.providerId = pointerize(m.provider_id)
     this.spark_version = pointerize(m.spark_version)
-    /** @type {import('./typings.js').FraudAssesment} */
-    this.fraudAssessment = null
+    /** @type {import('./typings.js').TaskingEvaluation} */
+    this.taskingEvaluation = null
+    /** @type {import('./typings.js').MajorityEvaluation} */
+    this.majorityEvaluation = null
```
juliangruber, line 28: Before we commit to this name: is `committeeEvaluation` a better name? I'm not sure myself

lib/retrieval-stats.js
```diff
@@ -81,8 +81,9 @@ export const buildRetrievalStats = (measurements, telemetryPoint) => {
     const endAt = m.end_at
     const ttfb = startAt && firstByteAt && (firstByteAt - startAt)
     const duration = startAt && endAt && (endAt - startAt)
+    const isAccepted = m.taskingEvaluation === 'OK' && m.majorityEvaluation === 'OK'
```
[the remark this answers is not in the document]
juliangruber: To double check: We will now be calling measurements that will be rewarded "accepted measurements", right?

lib/evaluate.js
juliangruber: Or should it rather be just "accepted measurements"?

```suggestion
    'EVALUATE ROUND %s: Evaluated %s measurements, accepted %s entries.\n%o',
```

juliangruber (changes requested):
Great work!

juliangruber:
Feel free to land this work next week during my vacation, without further review by me

lib/evaluate.js
bajtos: I like _rewarding %s entries_ 👍🏻

lib/platform-stats.js
bajtos: I'd rather not hide these two conditions in a helper function - we had that before with `fraudAssessment`, and it caused the introduction of subtle bugs.

The definition of what is an accepted measurement is different in different places.

We could use something like "m#rewarded", but again, I am concerned that we will tend to use it incorrectly in places that should treat non-majority measurements as accepted.

lib/preprocess.js
bajtos: I had the same idea!

But we already have `committee.evaluation` defined as `CommiteeEvaluation` here:

https://github.com/filecoin-station/spark-evaluate/blob/607d836f6a60b9b22c57254d6df041dcd4c1cf5f/lib/committee.js#L12-L18

It is a bit confusing.

There are two views on the data:

1. A list of measurements, where we look at `measurement.majorityEvaluation`. We are interested only in the evaluation result code (`OK`, `MINORITY_RESULT`, etc.)
2. A list of committees, each committee containing a list of measurements. We are interested in the full details of what the committee found (the CommiteeEvaluation object).

A possible path forward I see:

- Rename `measurement.majorityEvaluation` to `measurement.committeeEvaluation`, as you suggested.
- Rename `committee.evaluation` something like `committee.outcome`, `committee.result` or `committee.report`, rename the CommiteeEvaluation type accordingly.

WDYT?

lib/evaluate.js
bajtos: Good catch. This was introduced by https://github.com/filecoin-station/spark-evaluate/pull/439.

I'll remove this condition.

bajtos: See 1aba26794

lib/retrieval-stats.js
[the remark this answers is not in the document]
bajtos: Hmm, this is a good call! 🤔 

I want to call measurements that will be rewarded "measurementsToReward"; that's what I use inside the main evaluation loop.

Everywhere else, we need to decide whether we are interested in measurements that passed tasking evaluation but may be in the minority, or whether we are interested in measurements that will be rewarded (passed the tasking evalution and are in majority).

In that light, maybe it would be better to avoid using the term "accepted measurements" altogether. Or qualify it as "measurements accepted by tasking"?

I don't have a good answer 😞

NikolasHaimerl (commented):
LGTM, I'll approve once Julians changes have been addressed.

lib/preprocess.js
juliangruber: Great input! I think the bigger question is what an evaluation is supposed to be, and it would be nice if we didn't use that term for both the result code you attach to a measurement, and the detailed individual output.

Typing this out, what do you think about calling "evaluation" the string result code and "evaluationDetails" things like the committee evaluation result?

bajtos: I like `evaluationDetails` 👍🏻 

On second thought, I also prefer to keep `measurement.majorityEvaluation`. From the perspective of individual measurement, we are interested in knowing if the measurement is a majority view. 

Here are the possible values:

```js
majorityEvaluation = 'OK'
majorityEvaluation = 'MINORITY_RESULT'
majorityEvaluation = 'MAJORITY_NOT_FOUND'
majorityEvaluation = 'COMMITTEE_TOO_SMALL'
```

Let's compare that with using `committeeEvaluation`:

```js
committeeEvaluation = 'OK'
committeeEvaluation = 'MINORITY_RESULT'
committeeEvaluation = 'MAJORITY_NOT_FOUND'
committeeEvaluation = 'COMMITTEE_TOO_SMALL'
```

I propose renaming the `OK` value to `MAJORITY_RESULT` to make it more descriptive.

```js
majorityEvaluation = 'MAJORITY_RESULT'
majorityEvaluation = 'MINORITY_RESULT'
majorityEvaluation = 'MAJORITY_NOT_FOUND'
majorityEvaluation = 'COMMITTEE_TOO_SMALL'
```

@juliangruber What do you think?

If we decide to keep `majorityEvaluation` name, then I'd like to implement `evaluationDetails` renaming in a follow-up pull request.

juliangruber: Uh what about `consensusEvaluation`? Isn't that what it's really about?

bajtos: > Uh what about `consensusEvaluation`? Isn't that what it's really about?

Nice!

I renamed `majorityEvaluation` to `consensusEvaluation`, renamed `OK` to `MAJORITY_RESULT` and also renamed `committee.evaluation` to `commitee.evaluationDetails`.

See 7a5db2d

lib/committee.js
```diff
@@ -111,19 +111,22 @@ export class Committee {
 
     const retrievalResultMajority = this.#findMajority(retrievalResultProps)
     const hasRetrievalMajority = !!retrievalResultMajority
-    /** @type {CommitteeEvaluation['retrievalResult']} */
+    /** @type {CommitteeEvaluationDetails['retrievalResult']} */
```
juliangruber, line 114: I think this should be called `ConsensusEvaluationDetails`, to match `consensusEvaluation`

lib/typings.d.ts
```diff
@@ -34,15 +34,17 @@ export type CommitteeCheckError =
 | 'MAJORITY_NOT_FOUND'
 | 'MINORITY_RESULT'
 
+export type ConsensusEvaluation =
+  | 'MAJORITY_RESULT'
+  | CommitteeCheckError
```
juliangruber, line 39: should this be called `ConsensusCheckError`?

juliangruber (changes requested):

lib/committee.js
bajtos: The type `CommitteeEvaluationDetails` describes what's stored in `committee.evaluationDetails`.

lib/typings.d.ts
bajtos: Done in 85c882d

lib/committee.js
bajtos: Renamed in c9d91eb

lib/committee.js
```diff
@@ -3,18 +3,18 @@ import createDebug from 'debug'
...
+/** @import {RetrievalResult, ConsensusCheckError} from './typings.js' */
 
 const debug = createDebug('spark:committee')
 
 /** @typedef {Map<string, Committee>} TaskIdToCommitteeMap */
 
 /** @typedef {{
    hasIndexMajority: boolean;
-   indexerResult: string  | CommitteeCheckError;
+   indexerResult: string | ConsensusCheckError;
```
juliangruber, line 14: should this be called `ConsensusEvaluationError`? Otherwise we should call the other one `ConsensusCheckDetails`, to be consistent

juliangruber (changes requested):

lib/typings.d.ts
bajtos: I decided to use `ConsensusNotFoundReason`, see 3680486

juliangruber (approved):

bajtos:
@NikolasHaimerl would you like to review the final version or can I go ahead and land the PR?

test/evaluate.js
```diff
@@ -241,6 +241,43 @@ describe('evaluate', async function () {
     assertPointFieldValue(point, 'total_nodes', '3i')
   })
```
pyropy, line 243: Small nitpick -- should rename this file from `evaluate.js` to `evaluate.test.js` to make things consistent?

lib/preprocess.js
```diff
@@ -22,8 +22,10 @@ export class Measurement {
     // Note: providerId is recorded by spark-publish but we don't use it for evaluations yet
     this.providerId = pointerize(m.provider_id)
     this.spark_version = pointerize(m.spark_version)
-    /** @type {import('./typings.js').FraudAssesment} */
-    this.fraudAssessment = null
+    /** @type {import('./typings.js').TaskingEvaluation} */
```
pyropy, line 25: Since initial value is set to `null` by default we might should set property type to be optional, maybe something like:

```
/** @type {import('./typings.js').TaskingEvaluation?} */
```

NikolasHaimerl (approved):

lib/preprocess.js
bajtos: Agreed :+1:

This pull request is not making this any worse, so let's leave such improvement out of the scope, please.

pyropy (approved):

PR #450, opened 2025-01-20
Add publish script

pyropy:
Introduces new `publish` script to `package.json`. This publish script is used to publish the package using `np` package (same as it's done for filecoin station [core](https://github.com/filecoin-station/core)) 

Related to https://github.com/filecoin-station/spark-stats/issues/288


.github/workflows/publish.yml
```diff
@@ -0,0 +1,21 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+      # Setup .npmrc file to publish to npm
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
```
bajtos, line 16: Can we use v22 please? It's the latest LTS.

```suggestion
          node-version: 22
```

.github/workflows/publish.yml
```diff
@@ -0,0 +1,21 @@
...
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm publish --provenance --access public
```
bajtos, line 19: I believe you need to increase the package version before you can publish a new version to npm. npm does not allow you to overwrite the contents of an already published version.

That prompts a bigger question - how do you decide what is the next version number?

In the past, I had great experience with conventional commits (https://www.conventionalcommits.org/en/v1.0.0/), but that requires us to agree on this new process and add a CI step to ensure all commits are formatted according to conventional commits conventions.

Here is the setup we created in loopback-next, a project I used to maintain in the past:

https://github.com/loopbackio/loopback-next/blob/92fee0f161ccc2bb9d28579450f98fc43c959416/.github/workflows/continuous-integration.yml#L105-L119

https://github.com/loopbackio/loopback-next/blob/92fee0f161ccc2bb9d28579450f98fc43c959416/commitlint.config.js

bajtos (changes requested):

.github/workflows/publish.yml
pyropy: Sure thing. I guess that we also upgrade the CI workflow and Docker container base image to use node version 22 too at this point?

.github/workflows/publish.yml
juliangruber: In other modules that we need to publish to npm, we don't do that from npm - the dev simply runs `npm run publish`. Figuring out a proper CI publishing strategy is more complex. Can we publish manually for now?

juliangruber (changes requested):

.github/workflows/publish.yml
pyropy: Yes, let's go with manual publishing for now as @juliangruber suggested. I was not aware of the current process -- seems good enough.

package.json
```diff
@@ -1,14 +1,15 @@
 {
   "name": "spark-evaluate",
```
juliangruber, line 2:
```suggestion
  "name": "@filecoin-station/spark-evaluate",
```

README.md
```diff
@@ -169,3 +169,11 @@ node bin/dry-run.js \
 ```bash
 $ git push
 ```
+
+## Publish
+
+Publish a new version of `spark-evaluate` package:
```
juliangruber, line 175:
```suggestion
Publish a new version of `@filecoin-station/spark-evaluate`:
```

juliangruber (changes requested):

juliangruber (approved):
nice!

PR #475, opened 2025-02-13
Add HTTP HEAD success count. CheckerNetwork/spark#104

juliangruber:
For CheckerNetwork/spark#104
Picks up on https://github.com/CheckerNetwork/spark-api/pull/521. Neither PR is blocking the other, they can be merged in any order.
-> Blocks https://github.com/CheckerNetwork/spark-stats/pull/322


test/evaluate.js
```diff
@@ -98,7 +98,8 @@ describe('evaluate', async function () {
       total: 1,
       successful: 1,
       // None of the measurements use http
-      successful_http: 0
+      successful_http: 0,
+      successful_http_head: 0
```
pyropy, line 102: Could you add test cases to test the assert found in `assertValidMeasurement` ?

pyropy (changes requested):
Looking good so far👏🏻

lib/provider-retrieval-result-stats.js
```diff
@@ -16,17 +16,22 @@ const withPgClient = fn => async ({ createPgClient, ...args }) => {
...
       const minerId = m.minerId
-      const retrievalStats = providerRetrievalResultStats.get(minerId) ?? { total: 0, successful: 0, successfulHttp: 0 }
+      const retrievalStats = providerRetrievalResultStats.get(minerId) ?? { total: 0, successful: 0, successfulHttp: 0, successfulHttpHead: 0 }
       retrievalStats.total++
       if (m.retrievalResult === 'OK') {
         retrievalStats.successful++
-        if (m.protocol && m.protocol === 'http') { retrievalStats.successfulHttp++ }
+        if (m.protocol && m.protocol === 'http') {
+          retrievalStats.successfulHttp++
+          if (typeof m.head_status_code === 'number' && m.head_status_code >= 200 && m.head_status_code < 300) {
```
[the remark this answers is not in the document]
juliangruber: What do you mean by implement the status code correctly?

When a spark checker performs the HEAD request, the status code that the SP responds with is the thing we want to measure.

lib/preprocess.js
```diff
@@ -248,6 +249,7 @@ export const assertValidMeasurement = measurement => {
       'stationId must be a hex string with 88 characters'
     )
   }
+  assert(!(measurement.head_status_code && !measurement.status_code), '`head_status_code` must have `status_code` as well')
```
bajtos, line 252: To be safe, we should also assert that `head_status_code` is a number. WDYT?

test/provider-retrieval-result-stats.test.js
```diff
@@ -62,8 +63,8 @@ describe('Provider Retrieval Result Stats', () => {
         }
       ])
       assert.deepStrictEqual(stats, new Map([
-        ['0', { total: 2, successful: 2, successfulHttp: 1 }],
-        ['1', { total: 2, successful: 0, successfulHttp: 0 }]
+        ['0', { total: 2, successful: 2, successfulHttp: 1, successfulHttpHead: 1 }],
+        ['1', { total: 2, successful: 0, successfulHttp: 0, successfulHttpHead: 0 }]
```
bajtos, line 67: Hmm, this will be confusing.

IIUC, the test data have two measurements:
1. One measurement was reported by a checker node that tested HEAD retrieval and reported the result 200
2. The other measurement was reported by a checker node that did not perform any HEAD retrieval

IMO, Spark should report a 100% success rate of HEAD requests in such case. One HEAD request was performed, one HEAD request was successful.

Solution: add `totalHttpHead` in addition to `successfulHttpHead`.

```suggestion
        ['1', { total: 2, successful: 0, successfulHttp: 0, totalHttpHead: 1, successfulHttpHead: 0 }]
```

OTOH, one can argue that after all checkers upgrade to the latest Spark version, `totalHttpHead` should always be the same as `total`, so we don't need two columns.

However, a large portion of the checker network is slow to upgrade Spark checker code. To guarantee that all measurements include HEAD request result, we would have to start rejecting measurements without `headStatusCode`.

At yet other hand, one can also argue that we should keep these two columns, because if the "proper" retrieval check throws an error, then the checker node does not perform the head check, as can be seen here:

https://github.com/CheckerNetwork/spark-checker/pull/109/files#diff-9eb2d17c6352497045e4a22df4b4957d5d44f3cfff0c66c6519d07705ee1ed86R80-R83
```js
    await this.fetchCAR(provider.protocol, provider.address, retrieval.cid, stats)
    if (stats.protocol === 'http') {
      await this.testHeadRequest(provider.address, retrieval.cid, stats)
    }
```

Let's discuss!

bajtos (changes requested):

test/provider-retrieval-result-stats.test.js
bajtos: Spark Checker version break-down for the last 24 hours:

Screenshot 2025-02-17 at 15 24 30

lib/preprocess.js
juliangruber: We are currently not doing any type validation on measurements fields, eg I couldn't find any for `status_code` either. We're relying on `spark-publish` submitting this correctly (where we do have validation as well).

test/provider-retrieval-result-stats.test.js
juliangruber: > At yet other hand

I'm slowly running out of hands here!

> IIUC, the test data have two measurements:
>
> One measurement was reported by a checker node that tested HEAD retrieval and reported the result 200
The other measurement was reported by a checker node that did not perform any HEAD retrieval
IMO, Spark should report a 100% success rate of HEAD requests in such case. One HEAD request was performed, one HEAD request was successful.

In https://github.com/CheckerNetwork/spark-checker/issues/104 you suggested that HEAD requests are _only_ performed after successful retrievals. Therefore, the total number of head requests = the number of successful retrievals. So, we already have all the data and the reader can interpret as desired.


> OTOH, one can argue that after all checkers upgrade to the latest Spark version, totalHttpHead should always be the same as total, so we don't need two columns.

Exactly, that was my plan.

> However, a large portion of the checker network is slow to upgrade Spark checker code. To guarantee that all measurements include HEAD request result, we would have to start rejecting measurements without headStatusCode.

If checker don't upgrade, then we will get an artificially low head_success/head_total ratio, which is painting a false picture. Thank you for calling this out!

> At yet other hand, one can also argue that we should keep these two columns, because if the "proper" retrieval check throws an error, then the checker node does not perform the head check, as can be seen here:

I'm not following. What do you mean by "two columns" here, and what is the necessary change? In my understanding, if the retrieval fails, we won't attempt a HEAD request. This also means that if the retrieval success, we will attempt a HEAD request. Therefore, successful retrieval count is still the number of attempted HEAD requests (when the checker is up to date).

Re updates:
I propose we give a 7d upgrade window after shipping the updated checker version, after which we merge this one here. I'm going to add a requirement for measurements to have head_status_code property to a new spark-api PR.

lib/provider-retrieval-result-stats.js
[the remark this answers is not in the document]
juliangruber: ping @NikolasHaimerl

lib/preprocess.js
juliangruber: ping @bajtos

test/provider-retrieval-result-stats.test.js
juliangruber: ping @bajtos

lib/provider-retrieval-result-stats.js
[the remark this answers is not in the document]
NikolasHaimerl: The stat `retrievalStats.successfulHttpHead++` is updated under the condition that the SP  sets `head_status_code` to a value we expect. In this case this is `if (typeof m.head_status_code === 'number' && m.head_status_code >= 200 && m.head_status_code < 300)`. 
I was wondering about the assumption of whether we can expect SPs to choose the correct `head_status_code` between `200` and `300`.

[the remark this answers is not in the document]
juliangruber: If SPs respond with a bad status code, this is exactly what we want to find out here. The goal of this work is to see whether SPs respond to HEAD requests and whether the response of the HEAD request matches the response of the GET request. Through this, we can measure whether it's possible for users to issue a HEAD request in order to get a quick reading on whether an SP is going to also respond successfully to a full GET request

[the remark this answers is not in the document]
NikolasHaimerl: Ah I see, thanks for taking the time to explain.

lib/preprocess.js
bajtos: Fair enough 👍🏻

test/provider-retrieval-result-stats.test.js
bajtos: >In my understanding, if the retrieval fails, we won't attempt a HEAD request. This also means that if the retrieval success, we will attempt a HEAD request. Therefore, **successful retrieval count is still the number of attempted HEAD requests** (when the checker is up to date).


That's a great insight that I did not realise 👏🏻 

> Re updates:
> I propose we give a 7d upgrade window after shipping the updated checker version, after which we merge this one here. I'm going to add a requirement for measurements to have head_status_code property to a new spark-api PR.

SGTM.

I agree we don't need to add any more columns like `totalHttpHead`.

bajtos (approved):

pyropy (approved):

test/provider-retrieval-result-stats.test.js
juliangruber: Converting to a draft PR to prevent accidental merges

juliangruber: Screenshot 2025-02-28 at 13 24 59

The vast majority of the network is on the latest version, we can continue now

PR #479, opened 2025-02-17
Add include minority results in evaluate script output. Closes #458

### LINKED ISSUE #458: Include minority retrieval results in the output of evaluate-measurements.js script
In the docs page [Inspect recent measurements](https://docs.filspark.com/troubleshooting-miner-score#block-7cb238d9569f4d1cac87270d416a606d), we instruct SPs to fetch recent measurements for their MinerID and inspect the evaluated results to see the retrieval results reported by Spark checker nodes.

Right now, the script `evaluate-measurements.js` outputs only the results in the majority. That makes it impossible to understand why the RSR score is not 100%.

Let's fix the script to output all results that passed the tasking evaluation.

Dependency:

- https://github.com/filecoin-station/spark-evaluate/pull/442

Prior work:

- https://github.com/filecoin-station/spark-evaluate/pull/396

juliangruber:
Closes #458

# Before

```console
$ node bin/evaluate-measurements.js measurements-f010479.ndjson
[...]
Found 484 accepted measurements.
  OK                                       480        (99.17%)
  TIMEOUT                                  1          (0.2%)
  HOSTNAME_DNS_ERROR                       2          (0.41%)
  IPNI_ERROR_FETCH                         1          (0.2%)
[...]
```

☝️ Notice how minority results are included in the output, although they aren't actionable for the SP.

```console
$ head -n2 measurements-f010479.evaluation.txt
Timestamp                CID                                                                    Protocol   RetrievalResult
2025-02-16T20:22:15.307Z bafybeibhg66222vgfas5yc72hpnnydglezznfwjjwcgro6jgnj2st5ydi4            http       OK
```

☝️ Here there's no notion of whether something is a minority result or not.

```console
$ KEEP_REJECTED=1 node bin/evaluate-measurements.js measurements-f010479.ndjson
[...]
Found 484 accepted measurements.
  OK                                       480        (99.17%)
  TIMEOUT                                  1          (0.2%)
  HOSTNAME_DNS_ERROR                       2          (0.41%)
  IPNI_ERROR_FETCH                         1          (0.2%)
[...]
```

☝️ Same problem as above

```console
$ head -n2 measurements-f010479.evaluation.txt
Timestamp                CID                                                                    Protocol   🕵️   RetrievalResult
2025-02-16T20:22:15.307Z bafybeibhg66222vgfas5yc72hpnnydglezznfwjjwcgro6jgnj2st5ydi4            http       🫡   OK
```

☝️ It's not clear what 🕵️ means.

# After

```console
$ node bin/evaluate-measurements.js measurements-f010479.ndjson
Found 433 accepted measurements.
  OK                                       429        (99.07%)
  MINORITY_RESULT                          4          (0.92%)
```

☝️ See how the output has improved, and unactionable minority results are grouped in one row. Measurements that don't pass the tasking algorithm aren't included. Based on the above, the SP should have 100% RSR for the measurements tested. 

@bajtos should we exclude `MINORITY_RESULT` here and just say OK 100%?

```console
$ head -n2 measurements-f010479.evaluation.txt
Timestamp                CID                                                                    Protocol   Consensus RetrievalResult
2025-02-16T20:22:15.307Z bafybeibhg66222vgfas5yc72hpnnydglezznfwjjwcgro6jgnj2st5ydi4            http       🫡        OK
```

Notice new "Consensus" field. Measurements that don't pass the tasking algorithm aren't included.

```console
$ KEEP_REJECTED=1 node bin/evaluate-measurements.js measurements-f010479.ndjson
[...]
Found 484 accepted measurements.
  OK                                       429        (88.63%)
  TASK_WRONG_NODE                          23         (4.75%)
  MINORITY_RESULT                          4          (0.82%)
  DUP_INET_GROUP                           27         (5.57%)
  TASK_NOT_IN_ROUND                        1          (0.2%)
[...]
```

☝️ 👇  These now includes codes for everything: retrieval result, tasking failure and consensus failure.

```console
$ head -n2 measurements-f010479.evaluation.txt
Timestamp                CID                                                                    Protocol   Tasking Consensus RetrievalResult
2025-02-16T20:22:15.307Z bafybeibhg66222vgfas5yc72hpnnydglezznfwjjwcgro6jgnj2st5ydi4            http       🫡      🫡        OK
```

### DIFF
--- a/bin/evaluate-measurements.js
+++ b/bin/evaluate-measurements.js
@@ -42,7 +42,7 @@ const EVALUATION_NDJSON_FILE = `${basename(measurementsPath, '.ndjson')}.evaluat
 const evaluationTxtWriter = fs.createWriteStream(EVALUATION_TXT_FILE)
 const evaluationNdjsonWriter = fs.createWriteStream(EVALUATION_NDJSON_FILE)
 
-evaluationTxtWriter.write(formatHeader({ includeEvaluation: keepRejected }) + '\n')
+evaluationTxtWriter.write(formatHeader({ keepRejected }) + '\n')
 
 const resultCounts = {
   total: 0
@@ -97,27 +97,23 @@ async function processRound (roundIndex, measurements, resultCounts) {
     prepareProviderRetrievalResultStats: async () => {}
   })
 
-  for (const m of round.measurements) {
-    // FIXME: we should include non-majority measurements too
-    // See https://github.com/filecoin-station/spark-evaluate/pull/396
-    if (m.taskingEvaluation !== 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT') continue
-    resultCounts.total++
-    resultCounts[m.retrievalResult] = (resultCounts[m.retrievalResult] ?? 0) + 1
+  if (!keepRejected) {
+    round.measurements = round.measurements.filter(m => m.taskingEvaluation === 'OK')
   }
 
-  if (!keepRejected) {
-    round.measurements = round.measurements
-      // Keep accepted measurements only
-      // FIXME: we should include non-majority measurements too
-      // See https://github.com/filecoin-station/spark-evaluate/pull/396
-      .filter(m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT')
-      // Remove the taskingEvaluation and consensusEvaluation fields as all accepted measurements have the same value
-      .map(m => ({ ...m, taskingEvaluation: undefined, majorityEvaluation: undefined }))
+  for (const m of round.measurements) {
+    resultCounts.total++
+    const status = m.taskingEvaluation !== 'OK'
+      ? m.taskingEvaluation
+      : m.consensusEvaluation !== 'MAJORITY_RESULT'
+        ? m.consensusEvaluation
+        : m.retrievalResult
+    resultCounts[status] = (resultCounts[status] ?? 0) + 1
   }
 
   evaluationTxtWriter.write(
     round.measurements
-      .map(m => formatMeasurement(m, { includeEvaluation: keepRejected }) + '\n')
+      .map(m => formatMeasurement(m, { keepRejected }) + '\n')
       .join('')
   )
   evaluationNdjsonWriter.write(
@@ -148,21 +144,20 @@ function isFlagEnabled (envVarValue) {
 /**
  * @param {import('../lib/preprocess.js').Measurement} m
  * @param {object} options
- * @param {boolean} [options.includeEvaluation]
+ * @param {boolean} [options.keepRejected]
  */
-function formatMeasurement (m, { includeEvaluation } = {}) {
+function formatMeasurement (m, { keepRejected } = {}) {
   const fields = [
     new Date(m.finished_at).toISOString(),
     (m.cid ?? '').padEnd(70),
     (m.protocol ?? '').padEnd(10)
   ]
 
-  if (includeEvaluation) {
-    // FIXME: we should distinguish tasking and majority evaluation
-    // See https://github.com/filecoin-station/spark-evaluate/pull/396
-    fields.push((m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT' ? '🫡  ' : '🙅  '))
+  if (keepRejected) {
+    fields.push((m.taskingEvaluation === 'OK' ? '🫡' : '🙅').padEnd(7))
   }
 
+  fields.push((m.consensusEvaluation === 'MAJORITY_RESULT' ? '🫡' : '🙅').padEnd(9))
   fields.push((m.retrievalResult ?? ''))
 
   return fields.join(' ')
@@ -170,19 +165,20 @@ function formatMeasurement (m, { includeEvaluation } = {}) {
 
 /**
  * @param {object} options
- * @param {boolean} [options.includeEvaluation]
+ * @param {boolean} [options.keepRejected]
  */
-function formatHeader ({ includeEvaluation } = {}) {
+function formatHeader ({ keepRejected } = {}) {
   const fields = [
     'Timestamp'.padEnd(new Date().toISOString().length),
     'CID'.padEnd(70),
     'Protocol'.padEnd(10)
   ]
 
-  if (includeEvaluation) {
-    fields.push('🕵️  ')
+  if (keepRejected) {
+    fields.push('Tasking')
   }
 
+  fields.push('Consensus')
   fields.push('RetrievalResult')
 
   return fields.join(' ')


bin/evaluate-measurements.js
```diff
@@ -42,7 +42,7 @@ const EVALUATION_NDJSON_FILE = `${basename(measurementsPath, '.ndjson')}.evaluat
 const evaluationTxtWriter = fs.createWriteStream(EVALUATION_TXT_FILE)
 const evaluationNdjsonWriter = fs.createWriteStream(EVALUATION_NDJSON_FILE)
 
-evaluationTxtWriter.write(formatHeader({ includeEvaluation: keepRejected }) + '\n')
```
juliangruber, line 45: "includeEvaluation" didn't make sense since we are dealing with tasking and consensus result individually. Is it ok though to call a measurement that passed tasking but not consensus "accepted"/"not rejected"? Or are accepted measurements only the ones that pass tasking _and_ consensus, and we need a new name here: `includeTasking`

juliangruber:
> Right now, the script evaluate-measurements.js outputs only the results in the majority. That makes it impossible to understand why the RSR score is not 100%.

Thinking about this more, I might have misunderstood the goal. We were including non-majority results before (see PR description), and those are important, for the non-consensus based scores. What I'm not understanding then is what should actually be changed?

bajtos:
It's great to see a fresh view on this 👏🏻 

>  Notice how minority results are included in the output, although they aren't actionable for the SP.

When I created https://github.com/CheckerNetwork/spark-evaluate/issues/458, I understood that the script prints only measurements that passed the tasking evaluation and were in the majority - based on the following condition used to filter measurements to print:

```js
if (m.taskingEvaluation !== 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT') continue
```

Which made the script useless for troubleshooting why an SP had a RSR smaller than 100%.

However, now that I am reading that line again, I think there is a bug. This is what I probably wanted to write in https://github.com/CheckerNetwork/spark-evaluate/pull/442:

```js
if (!(m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT')) continue
```


> We were including non-majority results before (see PR description), and those are important, for the non-consensus based scores. What I'm not understanding then is what should actually be changed?


When I wrote that issue, I was expecting to change the condition shown above to the following plus update the output to include information about the committee consensus:

```js
if (m.taskingEvaluation !== 'OK') continue
```

See https://github.com/CheckerNetwork/spark-evaluate/pull/396 for the very first iteration on that.

Having written the above, I like your ideas on how to revamp the output of this script. Maybe we can implement them on top of the fix too?

bin/evaluate-measurements.js
```diff
@@ -148,41 +144,41 @@ function isFlagEnabled (envVarValue) {
...
     (m.cid ?? '').padEnd(70),
     (m.protocol ?? '').padEnd(10)
   ]
 
-  if (includeEvaluation) {
-    // FIXME: we should distinguish tasking and majority evaluation
-    // See https://github.com/filecoin-station/spark-evaluate/pull/396
-    fields.push((m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT' ? '🫡  ' : '🙅  '))
+  if (keepRejected) {
+    fields.push((m.taskingEvaluation === 'OK' ? '🫡' : '🙅').padEnd(7))
```
NikolasHaimerl, line 157: Do we include a --help command or similar which explains what emoji means what?

juliangruber:
@bajtos got it! I took me a bit to realize that this is never what we want:

```js
if (m.taskingEvaluation !== 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT') continue
```

So, you're proposing to change the line to:

```js
if (m.taskingEvaluation !== 'OK') continue
```

>  plus update the output to include information about the committee consensus

What do you mean by this? Would this be the final loop:

```js
for (const m of round.measurements) {
    if (m.taskingEvaluation !== 'OK') continue
    resultCounts.total++
    const status = m.consensusEvaluation !== 'MAJORITY_RESULT'
        ? m.consensusEvaluation
        : m.retrievalResult
    resultCounts[status] = (resultCounts[status] ?? 0) + 1
  }
```

bin/evaluate-measurements.js
pyropy: This + maybe change to more intuitive emojis like ✅ and ❌

juliangruber: Let's see if just changing the emojis is enough :)

pyropy:
> @bajtos got it! I took me a bit to realize that this is never what we want:
> 
> ```js
> if (m.taskingEvaluation !== 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT') continue
> ```
> 
> So, you're proposing to change the line to:
> 
> ```js
> if (m.taskingEvaluation !== 'OK') continue
> ```
> 
> > plus update the output to include information about the committee consensus
> 
> What do you mean by this? Would this be the final loop:
> 
> ```js
> for (const m of round.measurements) {
>     if (m.taskingEvaluation !== 'OK') continue
>     resultCounts.total++
>     const status = m.consensusEvaluation !== 'MAJORITY_RESULT'
>         ? m.consensusEvaluation
>         : m.retrievalResult
>     resultCounts[status] = (resultCounts[status] ?? 0) + 1
> }
> ```

In the #456 @bajtos mentions this:
> Let's fix the script to output all results that passed the tasking evaluation.

In that case I think the new loop where we skip measurements that have not passed tasking evaluation would be correct solution.

bin/evaluate-measurements.js
```diff
@@ -97,27 +97,23 @@ async function processRound (roundIndex, measurements, resultCounts) {
...
-    round.measurements = round.measurements
-      // Keep accepted measurements only
-      // FIXME: we should include non-majority measurements too
-      // See https://github.com/filecoin-station/spark-evaluate/pull/396
-      .filter(m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT')
-      // Remove the taskingEvaluation and consensusEvaluation fields as all accepted measurements have the same value
-      .map(m => ({ ...m, taskingEvaluation: undefined, majorityEvaluation: undefined }))
+  for (const m of round.measurements) {
+    resultCounts.total++
+    const status = m.taskingEvaluation !== 'OK'
```
pyropy, line 106: Maybe it's just me but this double ternary operator makes this really hard to read. In case we leave this loop logic do you think plain `if` statements would make this a bit more readable?

bajtos (approved):
LGTM but please resolve @pyropy's comment above.

NikolasHaimerl (approved):

bin/evaluate-measurements.js
juliangruber: cf85474

juliangruber:
@bajtos I'm confused: one one hand you approved this PR, on the other you +1ed the change suggestion in https://github.com/CheckerNetwork/spark-evaluate/pull/479#issuecomment-2665605422

bin/evaluate-measurements.js
```diff
@@ -97,27 +97,28 @@ async function processRound (roundIndex, measurements, resultCounts) {
...
+    let status
+    if (m.taskingEvaluation !== 'OK') {
+      status = m.taskingEvaluation
+    } else if (m.consensusEvaluation !== 'MAJORITY_RESULT') {
+      status = m.consensusEvaluation
+    } else {
+      status = m.retrievalResult
+    }
+
+    resultCounts[status] = (resultCounts[status] ?? 0) + 1
```
bajtos, line 116: After I posted my comment above and re-read your code, I realised the version presented here is a great way how to summarise the measurements when `KEEP_REJECTED=1` 👍🏻

bajtos (changes requested):
(See my comment above.)

bajtos:
> @bajtos I'm confused again 🙈
> 
> The first comment tells me that I should update the loop to the snippet listed. The second comment seems to tell me that the current version is better.

Sorry for that 🙈 

- KEEP_REJECTED is disabled - use the code from https://github.com/CheckerNetwork/spark-evaluate/pull/479#issuecomment-2748621006
- KEEP_REJECTED is enabled
  - The code from my comment will print different values for individual retrieval statuses, i.e. we will print "OK = all_successfull/all_measurements %", which is different from "OK = accepted_successful"/"accepted_measurements %" printed in the default mode. I think this may lead to confusion.
  - I think your proposal (https://github.com/CheckerNetwork/spark-evaluate/pull/479#discussion_r2010474085) will produce a better summary for KEEP_REJECTED mode.

I think we are hitting diminishing returns here and there is no need to worry about the summary for KEEP_REJECTED=1 mode until it becomes a source of confusion.

bin/evaluate-measurements.js
```diff
@@ -97,27 +97,22 @@ async function processRound (roundIndex, measurements, resultCounts) {
...
-  if (!keepRejected) {
-    round.measurements = round.measurements
-      // Keep accepted measurements only
-      // FIXME: we should include non-majority measurements too
-      // See https://github.com/filecoin-station/spark-evaluate/pull/396
-      .filter(m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT')
-      // Remove the taskingEvaluation and consensusEvaluation fields as all accepted measurements have the same value
-      .map(m => ({ ...m, taskingEvaluation: undefined, majorityEvaluation: undefined }))
+  for (const m of round.measurements) {
+    if (m.taskingEvaluation !== 'OK') continue
```
bajtos, line 105: In my previous comment, I had slightly different mental image of what's implemented. I assumed we would calculate the summary from all measurements, even the rejected ones, when KEEP_REJECTED=1.

I see that you are always excluding them from the summary, therefore we will print the same summary regardless of the value of the KEEP_REJECTED flag.

That works for me! 👏🏻

bajtos (approved):
:shipit:

PR #481, opened 2025-02-17
feat: per client rsr

NikolasHaimerl:
This PR aims at adding per-client stats for measurements to spark evaluate. 
The implementation plan for this PR can be found [here](https://github.com/CheckerNetwork/roadmap/issues/193#issuecomment-2656042186). 
Closes [#193](https://github.com/CheckerNetwork/roadmap/issues/193)


NikolasHaimerl:
@juliangruber  @bajtos @pyropy I am still working on a more complete test coverage but would like to get in some early feedback on the function creating and updating per client rsr scores.

migrations/023.do.add-per-client-rsr-table.sql
```diff
@@ -0,0 +1,8 @@
+CREATE TABLE daily_client_retrieval_stats (
+  client_id TEXT NOT NULL,
+  day DATE NOT NULL,
```
bajtos, line 3: Let's be consistent with the order of columns in the primary key and in the definitions of other "daily" tables.

```suggestion
  day DATE NOT NULL,
  client_id TEXT NOT NULL,
```

test/public-stats.test.js
```diff
@@ -29,7 +30,7 @@ describe('public-stats', () => {
     await pgClient.query('DELETE FROM indexer_query_stats')
     await pgClient.query('DELETE FROM daily_deals')
     await pgClient.query('DELETE FROM retrieval_timings')
-
+    await pgClient.query('DELETE FROM daily_client_retrieval_stats')
```
bajtos, line 33: Please preserve the empty line; it serves as a visual delimiter, making it easier for me to orientate in long blocks of code.

test/public-stats.test.js
```diff
@@ -627,6 +628,59 @@ describe('public-stats', () => {
...
+        assert(totalMeasurements > 0)
+        // F0 and F1 do not share any measurements and the expected number of measurments can be computed given the total number of measurements
+        const expectedF1TotalMeasurments = totalMeasurements - f0Stats[0].total
+        const { rows: f1Stats } = await pgClient.query(
+          "SELECT day::TEXT,client_id,total,successful,successful_http FROM daily_client_retrieval_stats WHERE client_id = 'f1client'"
+        )
+        // All measurments are successful and successful_http
+        assert.deepStrictEqual(f1Stats, [{ day: today, client_id: 'f1client', total: expectedF1TotalMeasurments, successful: expectedF1TotalMeasurments, successful_http: expectedF1TotalMeasurments }])
+      })
+    })
```
bajtos, line 683: I find this test difficult to understand because I have to run L634-L658 in my head to build a mental picture of what is the input for the aggregation pipeline.

Let's define test data explicitly as we do in other tests, please.

Also, let's find the minimum number of measurements required to reproduce the scenario you are testing. That should make the code defining `allMeasurements ` reasonably short.

bajtos (commented):
The implementation goes in the right direction 👍🏻

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
     JSON.stringify(rows)
   ])
 }
+
+/**
+ * @param {pg.Client} pgClient
+ * @param {Iterable<Committee>} committees
+ * @param {(minerId: string, cid: string) => (string[] | undefined)} findDealClients
+ */
+const updateDailyClientStats = async (pgClient, committees, findDealClients) => {
```
juliangruber, line 284:
```suggestion
const updateDailyClientRetrievalStats = async (pgClient, committees, findDealClients) => {
```

To be consistent with the line below and following

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+      console.warn(`Invalid retrieval task (${minerId}, ${payloadCid}): no deal clients found. Excluding the task from daily per-client stats.`)
+      Sentry.captureException(new Error('Invalid retrieval task: no deal clients found.'), {
+        extra: {
+          minerId,
+          payload_cid: payloadCid
+        }
+      })
+      continue
+    }
+    clients.forEach(clientId => {
```
juliangruber, line 302:
```suggestion
    for (const client of clients) {
```

Native `for..of` has superseded `Array#forEach` in most situations

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+
+/**
+ * @param {pg.Client} pgClient
+ * @param {Iterable<Committee>} committees
+ * @param {(minerId: string, cid: string) => (string[] | undefined)} findDealClients
+ */
+const updateDailyClientStats = async (pgClient, committees, findDealClients) => {
+  const clientRetrievalStats = new Map()
+  /** @type {Map<string, Array<Measurement>>} */
+  const clientToMeasurementsMap = new Map()
```
juliangruber, line 287:
```suggestion
  const clientMeasurements = new Map()
```

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+          payload_cid: payloadCid
+        }
+      })
+      continue
+    }
+    clients.forEach(clientId => {
+      if (clientToMeasurementsMap.has(clientId)) {
+        clientToMeasurementsMap.get(clientId).push(...c.measurements)
+      } else {
+        clientToMeasurementsMap.set(clientId, Array.from(c.measurements))
```
juliangruber, line 306:
```suggestion
        clientToMeasurementsMap.set(clientId, c.measurements)
```

`Committee#measurements` already is an array, see https://github.com/CheckerNetwork/spark-evaluate/blob/ba0ffe197774f7d19a298ecbfa4097109adf29c8/lib/committee.js#L20

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+      }
+    })
+  }
+
+  clientToMeasurementsMap.forEach((measurements, clientId) => {
```
juliangruber, line 311:
```suggestion
  for (const [clientId, measurements] of Object.entries(clientToMeasurementsMap)) {
```

See https://stackoverflow.com/a/54550693

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+    for (const m of measurements) {
+      const retrievalStats = clientRetrievalStats.get(clientId) ?? { total: 0, successful: 0, successfulHttp: 0 }
```
juliangruber, line 313:
```suggestion
    const retrievalStats = clientRetrievalStats.get(clientId) ?? { total: 0, successful: 0, successfulHttp: 0 }
    for (const m of measurements) {
```

This operation doesn't need to be repeated once per measurement, as it's the same for all per-client measurements

lib/public-stats.js
```diff
@@ -274,3 +275,77 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+      retrievalStats.total++
+      if (m.retrievalResult === 'OK') {
+        retrievalStats.successful++
+        if (m.protocol && m.protocol === 'http') { retrievalStats.successfulHttp++ }
+      }
+      clientRetrievalStats.set(clientId, retrievalStats)
```
juliangruber, line 319:
```suggestion
```

No need to replace the stats object since we have already updated the values It points to above (`#get()` returns a reference, like most JS APIs, and not a copy)

lib/public-stats.js
NikolasHaimerl: I am getting an error that says the object is not extensible if I do not wrap it in an Array.

test/public-stats.test.js
NikolasHaimerl: Thanks for the input. I simplified the test.

lib/public-stats.js
```diff
@@ -274,3 +275,79 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+    if (!clients || !clients.length) {
+      console.warn(`Invalid retrieval task (${minerId}, ${payloadCid}): no deal clients found. Excluding the task from daily per-client stats.`)
+      Sentry.captureException(new Error('Invalid retrieval task: no deal clients found.'), {
+        extra: {
+          minerId,
+          payload_cid: payloadCid
+        }
+      })
+      continue
+    }
```
pyropy, line 301: Small nitpick, but could you include one line after end of each `if` or `for` block. There are quite many of them given that we have to map data back to client and lack of whitespaces makes it a bit hard to read thought.

test/public-stats.test.js
```diff
@@ -627,6 +629,62 @@ describe('public-stats', () => {
...
+
+        // Separate the measurments into two groups, one for the client f0 and the other for f1
+        const findDealClients = (minerId, _cid) => {
+          switch (minerId) {
+            case 'f0test':
+              return ['f0client']
+            case 'f1test':
+              return ['f1client']
+            default:
+              throw new Error('Unexpected minerId')
```
pyropy, line 657: Is it possible for `findDealClients` to throw error? If so it would be good to include it the test case.

pyropy (commented):
Good job, I think PR is almost in ready state👏🏻

test/public-stats.test.js
NikolasHaimerl: It should not be possible no. I only included it so I know that something in my test setup is wrong.

test/public-stats.test.js
```diff
@@ -627,6 +629,198 @@ describe('public-stats', () => {
...
+          findDealClients
+        )
+        const { rows: f0Stats } = await pgClient.query(
+          "SELECT day::TEXT,client_id,total,successful,successful_http FROM daily_client_retrieval_stats WHERE client_id = 'f0client'"
+        )
+        assert.deepStrictEqual(f0Stats, [
+          { day: today, client_id: 'f0client', total: 2, successful: 2, successful_http: 2 }
+        ])
+        const { rows: f1Stats } = await pgClient.query(
+          "SELECT day::TEXT,client_id,total,successful,successful_http FROM daily_client_retrieval_stats WHERE client_id = 'f1client'"
```
pyropy, line 720:
```suggestion
        const { rows } = await pgClient.query(
          'SELECT day::TEXT, client_id, total, successful, successful_http FROM daily_client_retrieval_stats ORDER BY client_id'
```

You could do one select query here where you do full table scan and just do one assert.

test/public-stats.test.js
```diff
@@ -627,6 +629,198 @@ describe('public-stats', () => {
...
+        )
+        assert.deepStrictEqual(f1Stats, [
+          { day: today, client_id: 'f1client', total: 5, successful: 5, successful_http: 5 }
+        ])
```
pyropy, line 724:
```suggestion
        assert.deepStrictEqual(f1Stats, [
                  { day: today, client_id: 'f0client', total: 2, successful: 2, successful_http: 2 },
          { day: today, client_id: 'f1client', total: 5, successful: 5, successful_http: 5 }
        ])
```

Continuation of [previous comment](https://github.com/CheckerNetwork/spark-evaluate/pull/481/files#r1963014626)

test/public-stats.test.js
juliangruber: Can we annotate please that this should never happen? From looking at the code it's not clear whether this is an error that we can handle or whether this error shows a flaw in our system

test/public-stats.test.js
```diff
@@ -627,6 +629,198 @@ describe('public-stats', () => {
...
+          pgClient,
+          committees,
+          findDealClients
+        )
+        const { rows: f0Stats } = await pgClient.query(
+          "SELECT day::TEXT,client_id,total,successful,successful_http FROM daily_client_retrieval_stats WHERE client_id = 'f0client'"
+        )
+        assert.deepStrictEqual(f0Stats, [
+          { day: today, client_id: 'f0client', total: 2, successful: 2, successful_http: 2 }
+        ])
```
pyropy, line 718:
```suggestion
```
If you do only one full table scan you won't need two queries and two asserts.

pyropy (commented):
Really good job! 👏🏻 Let's make these tests a bit cleaner and I think we'll be ready to go.

lib/public-stats.js
```diff
@@ -274,3 +275,83 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+
+/**
+ * @param {pg.Client} pgClient
+ * @param {Iterable<Committee>} committees
+ * @param {(minerId: string, cid: string) => (string[] | undefined)} findDealClients
+ */
+export const updateDailyClientRetrievalStats = async (pgClient, committees, findDealClients) => {
+  const clientRetrievalStats = new Map()
+  /** @type {Map<string, Array<Measurement>>} */
+  const measurementsPerClient = new Map()
```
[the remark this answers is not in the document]
NikolasHaimerl: I should have made this clearer. I did change it to `clientMeasurements` but then I thought this does not quite seem right. It is not a client's measurements it is measurements that were done that are associated with a certain  client, which is why I renamed it. It is a good idea to also rename `clientRetrievalStats`.

test/public-stats.test.js
NikolasHaimerl: Does it make much of a difference whether two queries are done or one query and then a filter operation to separate f1client from f0client?

pyropy: > Does it make much of a difference whether two queries are done or one query

I don't think it makes much difference in regards to performance or the end result (testing if logic was valid), but it does make code cleaner and simpler.

> ... and then a filter operation to separate f1client from f0client?

In the suggested change I've provided you would not filter the result, rather you would test the whole response, for example:

```javascript
const { rows } = await pgClient.query(
`SELECT 
     day::TEXT, 
     client_id, 
     total, 
     successful, 
     successful_http 
  FROM daily_client_retrieval_stats 
  ORDER BY client_id`)

assert.deepStrictEqual(f1Stats, [
  { day: today, client_id: 'f0client', total: 2, successful: 2, successful_http: 2 },
  { day: today, client_id: 'f1client', total: 5, successful: 5, successful_http: 5 }
])
```

test/public-stats.test.js
pyropy: I see that I made an error in this code suggestion, `f1Stats` should actually be `rows`

lib/public-stats.js
```diff
@@ -274,3 +275,83 @@ const updateRetrievalTimings = async (pgClient, committees) => {
...
+      continue
+    }
+
+    // Each of the clients in the given list has a deal with the miner and payload CID of this committees retrieval task
+    // Each of the measurements in the committee is associated with every single one of these clients and should be accounted for in the stats
+    for (const clientId of clients) {
+      if (measurementsPerClient.has(clientId)) {
+        measurementsPerClient.get(clientId).push(...c.measurements)
+      } else {
+        measurementsPerClient.set(clientId, Array.from(c.measurements))
```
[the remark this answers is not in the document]
NikolasHaimerl: The error originates from [Object.freeze(ret)](https://github.com/CheckerNetwork/spark-evaluate/blob/4bdc217b35dcf2a42291cea74f224881c1c2eda9/lib/committee.js#L41). This line makes the measurement array non-extensible, leading to the following error:  

> **TypeError:** Cannot add property 2, object is not extensible  

To resolve this, you can use `Array.from(ret)`, which creates a new, mutable array from the frozen one, preventing the issue.

test/public-stats.test.js
NikolasHaimerl: Thanks for your input. I addressed it in bc27e9a

test/public-stats.test.js
NikolasHaimerl: I addressed this in bc27e9a

test/public-stats.test.js
NikolasHaimerl: I addressed this in bc27e9a

test/public-stats.test.js
```diff
@@ -627,6 +629,198 @@ describe('public-stats', () => {
...
+          committees,
+          findDealClients
+        )
+        const { rows: f0Stats } = await pgClient.query(
+          "SELECT day::TEXT,client_id,total,successful,successful_http FROM daily_client_retrieval_stats WHERE client_id = 'f0client'"
+        )
+        assert.strictEqual(f0Stats.length, 1)
+        assert.deepStrictEqual(f0Stats, [
+          { day: today, client_id: 'f0client', total: 2, successful: 2, successful_http: 2 }
+        ])
```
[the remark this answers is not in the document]
NikolasHaimerl: I addressed this in bc27e9a

lib/public-stats.js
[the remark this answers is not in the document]
NikolasHaimerl: I addressed this in c302a3f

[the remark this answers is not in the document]
bajtos: >  I did change it to `clientMeasurements` but then I thought this does not quite seem right. It is not a client's measurements it is measurements that were done that are associated with a certain client

FWIW, I agree with that - I would also expect `clientMeasurements` to contain measurements made by a specific client.

I like the new name `clientRetrievalStats`

lib/public-stats.js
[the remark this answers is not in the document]
bajtos: This TypeScript error has actually caught a real problem.

If we stored the original `c.measurements` array in the hashmap, then the next set of measurements for this client would push the new measurements to that array instance shared with the other committee. That would be a subtle bug.

An alternative solution that may be easier to reason about:

```js
let value = measurementsPerClient.get(clientId)
if (!value) {
  value = []
  measurementsPerClient.set(clientId, value)
}
value.push(...c.measurements)
```

test/public-stats.test.js
```diff
@@ -4,9 +4,10 @@ import pg from 'pg'
 import { DATABASE_URL } from '../lib/config.js'
 import { migrateWithPgClient } from '../lib/migrate.js'
 import { buildEvaluatedCommitteesFromMeasurements, VALID_MEASUREMENT } from './helpers/test-data.js'
-import { updatePublicStats } from '../lib/public-stats.js'
+import { updateDailyClientRetrievalStats, updatePublicStats } from '../lib/public-stats.js'
 import { beforeEach } from 'mocha'
 import { groupMeasurementsToCommittees } from '../lib/committee.js'
+import { describe } from 'node:test'
```
bajtos, line 10: We are using `mocha` in this repository. Mocha defines `describe` and `it` as global variables.

Please remove this line.

```suggestion
```

test/public-stats.test.js
```diff
@@ -627,6 +629,199 @@ describe('public-stats', () => {
         { day: today, miner_id: 'f1second', ttfb_p50: [1500] }
       ])
     })
+
+    describe('updateDailyClientRetrievalStats', () => {
+      it('aggregates per client stats', async () => {
+        // We create multiple measurments with different miner ids and thus key ids
+        // We also want to test multiple different number of measurments for a given combination of (cid,minerId)
```
bajtos, line 636: Typo:

```suggestion
        // We also want to test multiple different number of measurements for a given combination of (cid,minerId)
```

test/public-stats.test.js
```diff
@@ -627,6 +629,199 @@ describe('public-stats', () => {
...
+        }
+        await updateDailyClientRetrievalStats(
+          pgClient,
+          committees,
+          findDealClients
+        )
+        // Warning function should have been called
+        assert(warnCalled)
+        // Reset warning function
+        console.warn = originalWarn
```
bajtos, line 752: When the assertion above fails, an error is thrown and `console.warn` function is not restored.

You can solve this e.g. by moving this line above the assert line. That will still not work in case `updateDailyClientRetrievalStats` fails.

The bullet-proof solution is to save the original warn function in a `beforeEach` hook and restore it in `afterEach` hook.

test/public-stats.test.js
bajtos: This typo is repeated multiple times in the new code, please fix it everywhere.

test/public-stats.test.js
bajtos: If you want to fix this quickly, then it's IMO sufficient to swap the order asserting and restoring.

```js
// Reset warning function
console.warn = originalWarn

// Warning function should have been called
assert(warnCalled)
```

bajtos (commented):
The new version looks much better now! There are still a few details to improve; I'll leave it to others to do the final review & approval.

test/public-stats.test.js
NikolasHaimerl: Thanks, I addressed this in commit 807c7b5

lib/public-stats.js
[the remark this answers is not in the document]
NikolasHaimerl: Resolved in 3bc064f

juliangruber (approved):

juliangruber:
very nice frog gif

PR #488, opened 2025-02-26
fix: bug in parsing payload cid when updating daily per client rsr scores

NikolasHaimerl:
This PR fixes a bug that occurs while [passing the payload CID from the retrieval task list](https://github.com/CheckerNetwork/spark-evaluate/blob/daffe42466b67c483364b501f39c6d2dfbf2bd1d/lib/public-stats.js#L286).
The logs for the bug can be found [here](https://fly-metrics.net/d/fly-logs/fly-logs?orgId=150118&var-app=spark-evaluate).
Essentially, a wrong json field was used (`payloadCid` instead of `cid`) which resulted in `payloadCid` being `undefined` which causes the `findDealClients` to return an empty list of clients. This ultimately resulted in the warning [log](https://github.com/CheckerNetwork/spark-evaluate/blob/daffe42466b67c483364b501f39c6d2dfbf2bd1d/lib/public-stats.js#L289) seen in the logs.
I added an assertion during testing to verify the `payloadCid` when calling `findDealClients`.

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -10,8 +10,9 @@ jobs:
     runs-on: ubuntu-latest
     if: ${{ github.actor == 'dependabot[bot]' }}
     steps:
+      - name: Authenticate cli with a PAT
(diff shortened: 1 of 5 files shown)


pyropy (approved):
Good job on catching this 🐛

PR #243, opened 2024-05-31
feat: dump evaluated measurements after dry run

bajtos:
Modify the dry-run script to read the environment variable DUMP and save evaluated measurements to a CSV file when the env var is set.

The CSV file can be easily converted to a spreadsheet, which makes it easy to perform further data analysis.

**Example usage**

Dump all measurements:

```
❯ DUMP=1 node bin/dry-run.js 7970
(...lots of logs...)
Evaluated measurements saved to measurements-7970-all.csv
```

Dump measurements for miner with id `f0123`:

```
❯ DUMP=f0123 node bin/dry-run.js 7970
(...lots of logs...)
Storing measurements for miner id f0123
Evaluated measurements saved to measurements-7970-f0123.csv
```

Dump measurements submitted by participant with address `0xdead`:

```
❯ DUMP=0xdead node bin/dry-run.js 7970
(...lots of logs...)
Storing measurements from participant address 0xdead
Evaluated measurements saved to measurements-7970-0xdead.csv
```

PR #489, opened 2025-02-26
Auto-update Node.js version in Dockerfile

Hany-Almnaem:
Fixes[#243](https://github.com/CheckerNetwork/roadmap/issues/243)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,7 @@
 # syntax = docker/dockerfile:1
 
 # Adjust NODE_VERSION as desired
-ARG NODE_VERSION=22.13.1
-FROM node:${NODE_VERSION}-slim AS base
+FROM node:22.13.1-slim
 
 LABEL fly_launch_runtime="NodeJS"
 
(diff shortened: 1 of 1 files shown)


pyropy:
Thank you for your contribution @Hany-Almnaem 🎉 Could you update the description to link to the issue?

Dockerfile
```diff
@@ -1,8 +1,7 @@
 # syntax = docker/dockerfile:1
 
 # Adjust NODE_VERSION as desired
-ARG NODE_VERSION=22.13.1
-FROM node:${NODE_VERSION}-slim AS base
+FROM node:22.13.1-slim
```
pyropy, line 4:
```suggestion
FROM node:22.13.1-slim AS base
```

We're using multi-stage docker build in this Dockerfile, and removing the `AS <name>` causes build to break.

pyropy (changes requested):
Thank you for putting in great work 🙏🏻 

Let's make sure that build is passing before we merge this fix.

juliangruber (approved):

PR #498, opened 2025-03-10
fix: accept Graphsync measurements without HEAD status

bajtos:
Fix the regression introduced by e2a458b (#475)

- Graphsync retrievals don't test HEAD request, see https://github.com/CheckerNetwork/spark-checker/commit/3f8edc5539ddf83f8cb2cc1ccdd534a7a1624f1f
- As a result, measurements for Graphsync retrievals have `head_status_code` set to undefined or null.
- The validation performed during the preprocessing step discarded such measurements as invalid.

This pull requests relaxes the validation during preprocessing to assert numeric `head_status_code` for HTTP retrievals only.


NikolasHaimerl (approved):

bajtos:
After deploying this fix, Spark RSR jumped back to ~20%, where it was before Feb 28th.

Screenshot 2025-03-11 at 12 47 36

PR #512, opened 2025-04-02
fix: exlude IPNI_ERROR_5xx measurements from RSR

bajtos:
This PR has two commits to make the review easier:
- **refactor: fix test structure and names**
- **feat: exlude `IPNI_ERROR_5xx` measurements from RSR**

Links:
- https://github.com/CheckerNetwork/roadmap/issues/256

**feat: exlude `IPNI_ERROR_5xx` measurements from RSR**

Degradation of the IPNI service at cid.contact should not affect retrieval success rates.

This change is modifying the RSR calculations to exclude measurements with retrieval status `IPNI_ERROR_5xx` (e.g. `IPNI_ERROR_504`).

Important: we keep including `IPNI_ERROR_5xx` measurements in all other places. Most notably:

- Committees includes `IPNI_ERROR_5xx` measurements when looking for a majority consensus. This is important for the security of the protocol. If we excluded `IPNI_ERROR_5xx`s from the committee, then it would be easier for a malicious node to control the majority during IPNI outages, since the committees would be much smaller.

- RSR breakdowns submitted to InfluxDB and rendered in the Spark Internal Dashboard include `IPNI_ERROR_5xx` measurements. This is important to give us visibility into the data reported by the network.

- The RSR breakdown shown in the Spark public dashboard is based on the RSR breakdowns in InfluxDB, therefore it will show `IPNI_ERROR_5xx` as well.

**refactor: fix test structure and names**

I noticed that the tests for per-client and per-allocator RSRs are incorrectly nested inside a `describe()` block for a different feature.

I also noticed one test name refers to "aggregators" instead of "allocators".

This commit is fixing both problems.

The change is best reviewed in a diff ignoring white-space changes.

### DIFF
--- a/lib/provider-retrieval-result-stats.js
+++ b/lib/provider-retrieval-result-stats.js
@@ -21,6 +21,9 @@ export const build = committees => {
   for (const c of committees) {
     // IMPORTANT: include minority results in the calculation
     for (const m of c.measurements) {
+      // Ignore measurements reporting IPNI service outage
+      if (m.retrievalResult.match(/^IPNI_ERROR_5\d\d$/)) continue
+
       const minerId = m.minerId
       const retrievalStats = providerRetrievalResultStats.get(minerId) ?? { total: 0, successful: 0, successfulHttp: 0, successfulHttpHead: 0 }
       retrievalStats.total++
--- a/lib/public-stats.js
+++ b/lib/public-stats.js
@@ -390,6 +390,9 @@ function buildPerPartyStats (committees, perDealParty, partyName) {
     }
     const retrievalStats = retrievalStatsPerParty.get(partyId)
     for (const m of measurements) {
+      // Ignore measurements reporting IPNI service outage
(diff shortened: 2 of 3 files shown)


lib/provider-retrieval-result-stats.js
```diff
@@ -21,6 +21,9 @@ export const build = committees => {
   for (const c of committees) {
     // IMPORTANT: include minority results in the calculation
     for (const m of c.measurements) {
+      // Ignore measurements reporting IPNI service outage
+      if (m.retrievalResult.match(/^IPNI_ERROR_5\d\d$/)) continue
```
juliangruber, line 25:
```suggestion
      if (m.retrievalResult.startsWith('IPNI_ERROR_5')) continue
```

Simpler, but no big deal

juliangruber: Also applies below

juliangruber (approved):
Thank you for the thorough explanation and reasoning given in the PR title, that helped a lot 👏

PR #518, opened 2025-04-09
WIP: Evaluate network wide measurement

### LINKED ISSUE #517: Evaluate network wide retrievability metric
We need to add a new Spark score that measures how many deals (CIDs) can be retrieved from the network using any available retrieval provider, including non-Filecoin nodes running IPFS.

This new score is useful as a network-wide metric only. It must not affect the current per-miner/per-client/per-allocator RSR metrics. We shouldn't even collect it with per-miner/per-client/per-allocator granularity.

pyropy:
Closes #517

### DIFF
--- a/lib/preprocess.js
+++ b/lib/preprocess.js
@@ -17,6 +17,7 @@ export class Measurement {
   constructor (m, pointerize = (v) => v) {
     this.participantAddress = pointerize(parseParticipantAddress(m.participant_address))
     this.retrievalResult = pointerize(getRetrievalResult(m))
+    this.networkRetrievalResult = pointerize(getRetrievalResult(m))
     this.cid = pointerize(m.cid)
     this.minerId = pointerize(m.miner_id)
     // Note: providerId is recorded by spark-publish but we don't use it for evaluations yet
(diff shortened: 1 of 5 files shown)


lib/preprocess.js
```diff
@@ -41,6 +42,13 @@ export class Measurement {
     this.stationId = pointerize(m.station_id)
     this.carChecksum = pointerize(m.car_checksum)
     this.carTooLarge = m.car_too_large
+    this.networkRetrieval = {
+      statusCode: m.network_retrieval_status_code,
+      timeout: m.network_retrieval_timeout,
+      carTooLarge: m.network_retrieval_car_too_large,
+      endAt: m.network_retrieval_end_at,
```
bajtos, line 49:
```suggestion
      endAt: parseDateTime(m.network_retrieval_end_at),
```

lib/preprocess.js
```diff
@@ -41,6 +42,13 @@ export class Measurement {
     this.stationId = pointerize(m.station_id)
     this.carChecksum = pointerize(m.car_checksum)
     this.carTooLarge = m.car_too_large
+    this.networkRetrieval = {
```
bajtos, line 45: If we agree to use a different name than "networkRetrieval", then we need to rename this property too.

See https://github.com/CheckerNetwork/spark-checker/pull/132#discussion_r2037288495 and https://github.com/CheckerNetwork/spark-checker/pull/132#discussion_r2037309490

Plus, all other places that use the name "networkRetrieval" 😢

bajtos (changes requested):
The changes look reasonable, but we need to figure out the naming and measurement payload in other pull requests first.

juliangruber:
> The changes look reasonable, but we need to figure out the naming and measurement payload in other pull requests first.

+1 to figure out naming first.

Could you please link to the relevant PRs? I found https://github.com/CheckerNetwork/spark-checker/pull/132 in your comment

pyropy:
Should we evaluate alternative provider retrieval measurements separately using separate committees? There has been quite a few mentions of grouping measurements into committees in https://github.com/CheckerNetwork/spark-checker/pull/132.

bajtos:
I believe this PR is no longer relevant since we are shutting down the Leaderboard.

PR #520, opened 2025-04-11
feat: include minority results in telemetry stats

### LINKED ISSUE #446: Include non-majority measurements in retrieval result breakdown
Right now, we include majority measurements only. As a result, the retrieval-result-breakdown does not show outlier result types like infrequently-encountered HTTP error codes.

Ref: https://github.com/filecoin-station/spark-evaluate/pull/442#discussion_r1908899364

NikolasHaimerl:
Adds valid minority measurement results to the reported honest retrieval stats. 
Closes #446
Related to https://github.com/CheckerNetwork/spark-evaluate/pull/442#discussion_r1908899364

### DIFF
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -65,9 +65,6 @@ export const evaluate = async ({
   const measurementsToReward = measurements.filter(
     m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT'
   )
-  const validMinorityResults = measurements.filter(
-    m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MINORITY_RESULT'
-  )
 
   // Calculate reward shares
   const participants = {}
@@ -156,12 +153,13 @@ export const evaluate = async ({
   })
 
   const ignoredErrors = []
-
+  const honestMeasurements = measurements.filter(
+    m => m.taskingEvaluation === 'OK' && (m.consensusEvaluation === 'MINORITY_RESULT' || m.consensusEvaluation === 'MAJORITY_RESULT')
+  )
   try {
     recordTelemetry('retrieval_stats_honest', (point) => {
       point.intField('round_index', roundIndex)
-      buildRetrievalStats(measurementsToReward, point)
-      buildRetrievalStats(validMinorityResults, point)
+      buildRetrievalStats(honestMeasurements, point)
     })
   } catch (err) {
     console.error('Cannot record retrieval stats (honest).', err)
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -154,7 +154,7 @@ export const evaluate = async ({
 
   const ignoredErrors = []
   const honestMeasurements = measurements.filter(
-    m => m.taskingEvaluation === 'OK' && (m.consensusEvaluation === 'MINORITY_RESULT' || m.consensusEvaluation === 'MAJORITY_RESULT')
+    m => m.taskingEvaluation === 'OK'
   )
   try {
     recordTelemetry('retrieval_stats_honest', (point) => {
--- a/lib/evaluate.js
+++ b/lib/evaluate.js
@@ -65,6 +65,9 @@ export const evaluate = async ({
   const measurementsToReward = measurements.filter(
     m => m.taskingEvaluation === 'OK' && m.consensusEvaluation === 'MAJORITY_RESULT'
   )
+  const validMinorityResults = measurements.filter(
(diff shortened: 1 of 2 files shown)


lib/evaluate.js
```diff
@@ -153,13 +153,13 @@ export const evaluate = async ({
   })
 
   const ignoredErrors = []
-
+  const honestMeasurements = measurements.filter(
+    m => m.taskingEvaluation === 'OK' && (m.consensusEvaluation === 'MINORITY_RESULT' || m.consensusEvaluation === 'MAJORITY_RESULT')
```
bajtos, line 157: I can see how this condition is a literal implementation of what I described in https://github.com/CheckerNetwork/spark-evaluate/issues/446.

This will leave out outlier measurements where the committee was too small or no majority was found. That's not desired.

I would like to include all measurements accepted by the tasking evaluation.

```suggestion
    m => m.taskingEvaluation === 'OK'
```

lib/evaluate.js
```diff
@@ -153,13 +153,13 @@ export const evaluate = async ({
...
+  )
   try {
     recordTelemetry('retrieval_stats_honest', (point) => {
       point.intField('round_index', roundIndex)
-      // FIXME: Include non-majority measurements in these stats
-      // See https://github.com/filecoin-station/spark-evaluate/issues/446
-      buildRetrievalStats(measurementsToReward, point)
+      buildRetrievalStats(honestMeasurements, point)
```
bajtos, line 162: Please add some unit tests to verify the new implementation works as intended.

The tests (new and existing) should ideally cover all possible values of `m.consensusEvaluation`.

_(I am attaching this comment to a random source code line so that we can mark it as resolved later.)_

bajtos (changes requested):

lib/evaluate.js
NikolasHaimerl: Addressed in c6decc3

lib/evaluate.js
NikolasHaimerl: Addressed in https://github.com/CheckerNetwork/spark-evaluate/commit/c6decc311ca26bb31c77a83c3800618d09413142

test/retrieval-stats.test.js
```diff
@@ -324,6 +324,57 @@ describe('retrieval statistics', () => {
...
+        consensusEvaluation: 'MAJORITY_RESULT'
+      },
+      {
+        ...VALID_MEASUREMENT,
+        participantAddress: '0xfour',
+        retrievalResult: 'HTTP_500',
+        taskingEvaluation: 'OK',
+        consensusEvaluation: 'MINORITY_RESULT'
+      }
+    ]
```
bajtos, line 364: I think it's enough to have one measurement with `consensusEvaluation: 'MAJORITY_RESULT'`. What benefits do you see when there are multiple such measurements in the input?

Let's also test measurements with consensus evaluation values `COMMITTEE_TOO_SMALL` and `MAJORITY_NOT_FOUND` to get the full coverage.

lib/evaluate.js
```diff
@@ -153,13 +153,13 @@ export const evaluate = async ({
   })
 
   const ignoredErrors = []
-
+  const honestMeasurements = measurements.filter(
```
bajtos, line 156: We are trying to move away from "honest", because even honest nodes can end up with their measurements rejected, e.g. when there are more than one Checker node running in the same IPv4 /24 subnet.

`acceptedMeasurements` is a better option.

Something like `measurementsAcceptedByTaskingEvaluation` would be more precise, but probably too verbose.

bajtos (changes requested):
Much better 👍🏻

test/retrieval-stats.test.js
NikolasHaimerl: I included an additional test in 1028726

lib/evaluate.js
NikolasHaimerl: Addressed in 1028726

bajtos: > I see, this might require a change in this line too then:
> 
> https://github.com/CheckerNetwork/spark-evaluate/blob/5c53e4baaa4979484dbee6e37b7e85d7964383c8/lib/evaluate.js#L160-L161

Yeah, it would be nice to change the data point name as well. But then we would have to update all places reading these data points from InfluxDB. I think it's not worth our time now.

NikolasHaimerl: Alright, I was thinking the same thing. All the dashboards that use this data would need to be changed.

